PR #137, opened 2016-06-21
RDoc-546 [studio] describe hot spare

iftahbe:

PR #138, opened 2016-06-21
RDoc-476 [Licensing] HotSpare license support article

iftahbe:


arekpalinski:
@iftahbe when you add a new article page then you also need to link it to the menu to make it visible there. You need to add the appropriate record to `.docs.json` file, in this case to the following one: https://github.com/ravendb/docs/blob/master/Documentation/3.5/Raven.Documentation.Pages/server/configuration/.docs.json

The same applies to https://github.com/ravendb/docs/pull/137

ppekrol:
@iftahbe Also please use 'Related articles' convention to link articles between each other like in this one: https://github.com/ravendb/docs/blob/master/Documentation/3.0/Raven.Documentation.Pages/indexes/creating-and-deploying.markdown

iftahbe:
Thanks

Documentation/3.5/Raven.Documentation.Pages/server/configuration/configuration-hot-spare-license.markdown
```diff
@@ -3,21 +3,25 @@
...
\ No newline at end of file
+{NOTE: Remarks}
+- Server can be only be used as a replication destination (master-slave scenario) <br >
+- There is a one-time option to activate the license, which will enable all features for a period of 96 hours <br >
+- The license includes an option of testing the activiation for a short period, for operational testing <br >
+{NOTE/}
+
+## Related articles
+
+- [activating the hot spare](../../studio/management/hot-spare)
```
arekpalinski: Can you please apply the convention we have in different articles that we also indicate a name of a section of a related article when it's different than the current one, like here:

https://github.com/ravendb/docs/blob/master/Documentation/3.5/Raven.Documentation.Pages/file-system/server/bundles/encryption.markdown

Also I think the name of the related article should start with upper case letter: 
- Studio : Activating the Hot Spare

PR #139, opened 2016-06-22
RDoc-544 and RDoc-533

efratshenhar:


Documentation/3.5/Raven.Documentation.Pages/start/whats-new.markdown
```diff
@@ -21,7 +21,6 @@
 - `[Configuration]` Exposed `Raven/TempPath` setting,
 - `[Configuration]` Exposed `Raven/Indexing/MaxNumberOfStoredIndexingBatchInfoElements` setting,
 - `[Configuration]` Exposed, Raven/MaxClauseCount setting,
-- `[Endpoint]` Added `/debug/indexing-perf-stats and `/debug/indexing-perf-stats-with-timings`, statistics (/stats) no longer contain indexing performance data,
 - `[Endpoint]` Added `/admin/detailed-storage-breakdown`,
```
arekpalinski: Yes, if they are already available in 3.0 we can't mention about them on `What's new` page. However we still need to have them documented in both 3.x versions in the following articles:
https://ravendb.net/docs/article-page/3.0/csharp/server/troubleshooting/debug-endpoints

Documentation/3.5/Raven.Documentation.Pages/studio/overview/tasks/subscriptions.markdown
```diff
@@ -0,0 +1,11 @@
+﻿# Tasks : Subscriptions
+
+This view lists all available subscriptions and the etag of the last acknowledged and processed document.
+You have the ability to manually set the last etag.
+
+![Figure 1. Tasks. Subscriptions.](images/tasks-subscription-1.png)
+
+## Related articles
+
+- [What are Data Subscriptions?](../../../client-api/data-subscriptions/what-are-data-subscriptions)
```
arekpalinski: We've got the convention that when we refer an article which is in a different section than the current one we apply the section name as well here, in that case:
-  Client API : What are Data Subscriptions?

Documentation/3.5/Raven.Documentation.Pages/start/whats-new.markdown
arekpalinski: Just checked that `/debug/indexing-perf-stats-with-timings` is not available in 3.0

efratshenhar: /debug/indexing-perf-stats in 3.0 is the same as /debug/indexing-perf-stats-with-timings in 3.5.
 /debug/indexing-perf-stats in 3.5 is new.
Write both of them as added in whats-new?

arekpalinski: I think we need to write there exactly what we did:
- /debug/indexing-perf-stats -> moved to /debug/indexing-perf-stats-with-timings
- /debug/indexing-perf-stats -> gives a new output

PR #141, opened 2016-06-23
RDoc-540 

efratshenhar:


ppekrol:
I do not entirely follow which PRs those commits refer to. Can you send another one (or force-update this one) with commit messages that contain the issue number (e.g. 'RDoc-123 added /studio/overview/status/debug').

Please do not forget to paste the PR link on our YouTrack.

efratshenhar:
hi pawel 
there were no references to fix but i found bad links and fixed them.
about what you wrote in YouTrack, i will check it right now.

efratshenhar:
I thought the intention in the issue was about links to 3.0 , but now i see you meant to all the places 3.0 is mentioned.
i will check it and add it to the PR

PR #142, opened 2016-06-26
RDoc-547 Document index rename feature

iftahbe:


arekpalinski:
The description looks good, however I would consider the location of it. In the studio section we mostly have an article per view. Here, we've just added an additional functionality to Index Edit view:

https://ravendb.net/docs/article-page/3.0/csharp/studio/overview/indexes/index-edit-view

So maybe put it there as a new paragraph. What do you think? Also the description of `Save` button should reference the desc of the rename feature there (even if it is be on the same page).

arekpalinski:
Just noticed that I merged this one because both commits were in https://github.com/ravendb/docs/pull/143. Please apply changes in a new PR.

PR #153, opened 2016-07-06
RDoc-494 [Studio] Added the ability to enable replication on an exist…

efratshenhar:
…ing database


Documentation/3.5/Raven.Documentation.Pages/studio/overview/settings/replication.markdown
```diff
@@ -57,6 +57,14 @@ Please refer to following [article](../../../client-api/bundles/how-client-integ
 
 {PANEL/}
 
+{NOTE:Note}
+
+In earlier version of RavenDB, enabling the replication bundle wasn't possible for an existing databases. Now it is.
```
arekpalinski: We should rather say that it's available since 3.5 version. Also need to check and describe what are implications of enabling it when we already have documents in a database - in particular what about replication history of document metadata.

efratshenhar: I discussed this with Michael and we thought it's best to just say not supported without describing the implications

arekpalinski: Does it mean we provide users with option to enable replication but we don't support that? 

I've just tried to enable replication with existing docs on the live-test, successfully replicated docs to another db. Then configured it to master-master mode, introduced some changed on both nodes and still worked fine.

efratshenhar: I think the meaning of not supported is "not officialy supported" and that the risk of using it is on the user

arekpalinski: I looked for the relevant tickets for that and found two:
http://issues.hibernatingrhinos.com/issue/RavenDB-2273
http://issues.hibernatingrhinos.com/issue/RavenDB-3640

It seems we just exposed it in 3.5 studio.

Indeed we do not recommend using this if docs are already in db. Please mention about possible conflicts in the warning message you have added so a user will be aware of consequences.

Documentation/3.5/Raven.Documentation.Pages/studio/overview/settings/replication.markdown
```diff
@@ -67,7 +67,9 @@ Since version 3.5, enabling the replication bundle is possible for an existing d
 
 {WARNING:Warning}
 
-This is not recommended if your database already contain documents.
+This is not recommended if your database already contain documents.   
+In case you decided to do it anyway, there will be a side effect where replication info will be missing from the document's metadata
+until you "touch" the document again.
```
arekpalinski: Please mention that missing replication specific metadata can cause conflicts - that's the most important info for the user.

PR #165, opened 2016-07-11
RDoc-498 [Studio] Added disk space verification before importing data…

efratshenhar:
… (DB and FS)

### DIFF
--- a/Documentation/3.5/Raven.Documentation.Pages/file-system/studio/tasks/export-and-import-views.markdown
+++ b/Documentation/3.5/Raven.Documentation.Pages/file-system/studio/tasks/export-and-import-views.markdown
@@ -18,6 +18,7 @@ The import follows the same procedure as the export. You can click `Choose file`
 
 {NOTE:Note}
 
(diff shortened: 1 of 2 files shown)


Documentation/3.5/Raven.Documentation.Pages/file-system/studio/tasks/export-and-import-views.markdown
```diff
@@ -18,6 +18,7 @@ The import follows the same procedure as the export. You can click `Choose file`
 
 {NOTE:Note}
 
-Since version 3.5, you can disable versioning bundle during import.
+- Since version 3.5, you can disable versioning bundle during import.
+- The message "No sufficient diskspace for import, consider using Raven.Smuggler.exe directly." might occur.
```
arekpalinski: What if we change the error message in RavenDB? No one will remember to change that here. I think would be better to place a more generic information. For example by adding a new {NOTE: Disk space verification} panel and inform that we check for free space before the actual import operation.

PR #174, opened 2016-07-20
RDoc-561 [studio] Internal Breakdown / slow doc counts

iftahbe:


arekpalinski:
Try to avoid high resolution screenshots like this one in the future:
https://github.com/iftahbe/docs/blob/5c1ab8678b557bf3452472f99ccc28f4990fadc0/Documentation/3.5/Raven.Documentation.Pages/studio/overview/status/images/status-storage-2.png

where most of the image is empty space - readability of text will decline if we scale down the image on ravendb.net page.

PR #176, opened 2016-07-21
RDoc-483 [Studio] Added AdministratorJS Console

efratshenhar:


arekpalinski:
We need to elaborate more on that feature - it's a really nice one. It gives a lot of administration power, allows to run RavenDB code, on a live instance directly from the studio. Might be extremely useful in support cases where we need to figure out what's going on a production system.

Please explain what's `database` field in the script (instance of our main class - `DocumentDatabase`)

https://ayende.com/blog/173896/ravendb-3-5-whirl-wind-tour-im-the-admin-and-i-got-the-power
http://issues.hibernatingrhinos.com/issue/RavenDB-3390

PR #178, opened 2016-07-21
RDoc-563 Tombstones will not be filtered when replicating using colle…

iftahbe:
…ction replication


Documentation/3.5/Raven.Documentation.Pages/studio/overview/settings/etl.markdown
```diff
@@ -24,6 +24,7 @@ There is an option to write a JS function [(examples)](../../../client-api/comma
 for each collection/destination, which means that the script will run on every replicated document from that collection.
 
 ![Figure 4. Settings. ETL. Defining Transformation and Filtering scripts.](images/settings_etl-4.png)
+{NOTE The deletion of documents (tombstones) will be replicated to their destination without applying the transformation script on them. /}
```
arekpalinski: Please underline that because tombstones won't have the script applied (we cannot do that because tombstone doesn't have a document's body), they won't be filtered out - meaning all deletes will be processed as usual.

PR #195, opened 2016-08-04
RDoc-478 [Studio] Refresh Studio screenshots (Unified L&F)

efratshenhar:
With RDoc-477 Global Configuration


Documentation/3.5/Raven.Documentation.Pages/studio/management/global-configuration.markdown
```diff
@@ -0,0 +1,14 @@
+﻿# Manage Your Server : Global Configuration
+This feature allows you to define the behavior of the system once, and have it apply to all your databases, 
+and using the RavenDB Cluster option, you can apply it across all you nodes in on go. 
+You can configure global behavior and override it on individual databases.
```
ppekrol: Maybe substitute this with:
"Global Configuration feature allows you to define behavior for all server databases in one place. This behavior is inherited automatically but also can be overridden by a user per database basis."

Documentation/3.5/Raven.Documentation.Pages/studio/management/global-configuration.markdown
```diff
@@ -0,0 +1,14 @@
...
+
```
ppekrol: Let's stress out that this is used by Clustering maybe? What do you think?
E.g.
"{NOTE}

Global Configuration is used extensively by a Clustering feature (link?) and will allow you to spread configuration across all nodes.

{NOTE/}"

BTW, can you confirm that feature? I do not recall this.

arekpalinski:
@efratshenhar just to be sure - did you verify the feature that Pawel asked about? For example, if I configure Versioning bundle there and I'm in a cluster, will it propagate to all nodes in a cluster?

efratshenhar:
Yes , I changed the Raven/Gloval/Versioning (as you can see in the picture) and then i saw it in all the nodes.

arekpalinski:
Yes, but bundles aren't configured by the settings that you can define in `Cluster-wide database settings` but through specific documents which are put as system docs in a database (names of system docs starts with `Raven/`). Setting `Raven/Global/Versioning` as you presented in the picture won't take any effect for Versioning bundle because that's the prefix of _a document_ not a database configuration setting (we need to change that screenshot), see here and look for its usages:

https://github.com/ayende/ravendb/blob/v3.5/Raven.Abstractions/Data/Constants.cs#L448

The question is: will the versioning configuration (setup by `Global configuration` -> `Versioning`) be propagated within a cluster? (will the document with prefix `Raven/Global/Versioning/` be propagated to all nodes)

efratshenhar:
For your question: No, the Raven/Global/Versioning/ is not propagated if we setup by Global configuration -> Versioning.
you said that setting Raven/Global/Versioning as in the picture won't take any effect.... so what exactly do I need to write in configuration key to make it work? Or a better question: what am I allowed to write in configuration key?

arekpalinski:
You can specify there a _valid_ configuration option:
https://ravendb.net/docs/article-page/3.0/Csharp/server/configuration/configuration-options

`Raven/Global/Versioning` is not a configuration option.

arekpalinski:
Since the configuration of bundles is not propagated in a cluster let us do it this way: 

- create two sections in `Global Configuration` article: `Bundles` and `Cluster-wide database settings`
- the first one will contain: "You can configure global behavior for the following bundles: .... " as you wrote
- the second one should describe the you can define database settings, you need to remember about db restart etc and underline there that they will be propagated within a cluster

This way we will avoid the confusion about what is propagated withing a cluster and what isn't.

efratshenhar:
There is some problem with cluster- wide database setting. For now I removed it from global configuration page. I will add this information after it will work fine ( Issue RDoc-566 Document cluster-aware global database settings).

Documentation/3.5/Raven.Documentation.Pages/studio/management/global-configuration.markdown
```diff
@@ -1,4 +1,7 @@
 ﻿# Manage Your Server : Global Configuration
+
+### bundles
```
arekpalinski: bundles -> Bundles

arekpalinski:
Okay. Let's add it later.

PR #203, opened 2016-08-08
RDoc-521 update indexDefinition propriety

IdanHaim:


arekpalinski:
That's all? No info about performance enhancement. @ml054 can you have a look?

arekpalinski:
It seems that the warning about `Increased indexing time` on https://ravendb.net/docs/article-page/3.0/Csharp/indexes/querying/suggestions should be updated. I'm not familiar with details however at least provide info that we are faster than in 3.0

ayende:
The difference is that in 3.0 we are generating a dictionary for suggestions per  distance & accuracy pair.
In 3.5, we are able to have a single dictionary for suggestions globally, and apply the distnace & accurancy per query

arekpalinski:
@ayende Just talked to @ml054 about this again and as I understand that changes has nothing to do with performance. Is it correct?

ayende:
It does, because we don't need to create mulitple dictionaries

*Hibernating Rhinos Ltd  *

Oren Eini* l CEO l *Mobile: + 972-52-548-6969

Office: +972-4-622-7811 *l *Fax: +972-153-4-622-7811

ml054:
Actually before change we created dictionary with field name as a key: ex. LastName and {Levenstein; 0.4} as value. During query looked for suggestions in folder: FieldName + Distance + Accuracy. If not found we could not return suggestions. 

Right now we use only field name as directory name and can query for any distance + accuracy. 

When comes to comparing performance in both old and new version we can create only single suggestions for field (in old version key was field name so we can't create multiple entries with different distance/accuracy). 

The only change that I can see it fact that right now we can use any distance / accuracy combination when queries - we don't have to choose them during indexing process.

ayende:
Yes, but didn't we create additional dictionaries on the fly if requested
by the query?

*Hibernating Rhinos Ltd  *

Oren Eini* l CEO l *Mobile: + 972-52-548-6969

Office: +972-4-622-7811 *l *Fax: +972-153-4-622-7811



On Tue, Aug 9, 2016 at 1:25 PM, Marcin Lewandowski <notifications@github.com

ml054:
I don't think so, we simply read from different fs dictionary. I omit performance improvement related to lack of string concat code (FieldName + Distance + Accuracy).

ayende:
okay

*Hibernating Rhinos Ltd  *

Oren Eini* l CEO l *Mobile: + 972-52-548-6969

Office: +972-4-622-7811 *l *Fax: +972-153-4-622-7811



On Tue, Aug 9, 2016 at 1:52 PM, Marcin Lewandowski <notifications@github.com

IdanHaim:
so do we still get high CPU usage when we use it?

ml054:
For me: if it had high cpu it will still have the same.

IdanHaim:
OK so I think this doc update can be merged without any more update

arekpalinski:
Please drop the information about improved suggestions performance from What's new page (include it in this PR)

PR #231, opened 2016-09-05
RDoc-507[RavenFS] Unified FS creation method, now it matches the DB one

iftahbe:


arekpalinski:
It seems that the following sample should be fixed (need full key = `Raven/FileSystems/NorthwindFS`):
https://github.com/iftahbe/docs/blob/3fa60ae7eccb7d43ad893240da822afead2372e0/Documentation/3.5/Samples/csharp/Raven.Documentation.Samples/FileSystem/ClientApi/Commands/Admin.cs#L66

based on test adjustments made here: https://github.com/ayende/ravendb/commit/a52468d2a16dc2337aa812c76b08e0f29e02473f

Also it might be worth mentioning about added `CreateFileSystemDocument` that creates a configuration document with default values

PR #237, opened 2016-10-21
Fix note about using Lucene Parser

krivulcik:


krivulcik:
Sorry, I got confused in configuring the server.

PR #262, opened 2017-03-17
:wrench: Fix PatchRequest type when adding a new Property

PureKrome:
- Was incorrectly suggesting to use the `PatchCommandType.Add` type when trying to add a simple primitive value to an existing document. This is used for adding items to a collection.


ppekrol:
Thanks!

PR #269, opened 2017-05-15
Typo whatever rather than whether

tmpreston:
See http://www.grammar-monster.com/easily_confused/weather_whether_wether.htm


ayende:
Thanks

PR #289, opened 2017-09-29
[RDoc-693] Fixed a few typos in the 'Working with document IDs' page (3.0 & 3.5)

sputier:
The "Working with document identifiers" pages (3.0 & 3.5) each have two typos.

Convenions.FindIdentityProperty => Conventions.FindIdentityProperty
Note the we used `companies/` as the prefix => Note that we used `companies/` as the prefix


ayende:
Thank you very much for the fixes.

PR #313, opened 2017-11-15
RDoc-760, RDoc-767, RDoc-768 playground-server, creating-document-store and what-is-a-document-store

IdanHaim:


Documentation/4.0/Raven.Documentation.Pages/client-api/creating-document-store.markdown
```diff
@@ -0,0 +1,24 @@
+﻿# How to create a document store?
+
+To create an instance of the `DocumentStore` you need to specify a list of URL addresses that points to RavenDB server nodes. Each URL must contain the TCP port on which the server is listening (the default port number is 8080).
```
ayende: Not true. We don't need to say anything about the port. Can assume users know how URLs work

IdanHaim: I will delete Each URL must contain the TCP port on which the server is listening (the default port number is 8080).

Documentation/4.0/Raven.Documentation.Pages/client-api/creating-document-store.markdown
```diff
@@ -0,0 +1,24 @@
...
+
+{CODE document_store_creation@ClientApi\CreatingDocumentStore.cs /}
+
+This will instantiate a communication channel between your application and the local RavenDB server instance.
+
+##Initialization and disposal
+
+The above code contains two things you should pay special attention. The first one is that you have to call `Initialize` method on the `DocumentStore` to get the fully initialized instance of `IDocumentStore` that you will be able to work with.
+The second thing you should note is that the whole `DocumentStore` initialization code is placed inside the `using` statement to ensure the object disposal. This is needed to guarantee a proper cleanup. In a real case scenario it should be done when an application shuts down.
```
ayende: I'm not sure that telling users to do disposal right at the start is a good idea.
We better first talk about singleton and then add a note that if they use more than once, they should dispose it.
If they use one per application, there is no need, the process shutdown will cleanup everything.

Documentation/4.0/Raven.Documentation.Pages/client-api/what-is-a-document-store.dotnet.markdown
```diff
@@ -0,0 +1,22 @@
...
+It acts as the connection manager and also exposes methods to perform all operations that you can run against an associated server instance.
+
+The document store object has a list of URL addresses that points to RavenDB server nodes.
+
+* `DocumentStore` acts against a remote server via HTTP requests, implement common `IDocumentStore` interface
+
+The document store ensures access to the following client API features:
+
+* [Session](../client-api/session/what-is-a-session-and-how-does-it-work)
+* [Commands](../client-api/commands/what-are-commands)
```
ayende: That is no longer the case, IIRC

Documentation/4.0/Raven.Documentation.Pages/start/playground-server.markdown
```diff
@@ -0,0 +1,21 @@
...
+
+- server accepts remote connectivity, so you can build your applications locally for testing purposes and specify its address - `http://4.live-test.ravendb.net/` (more [here](../client-api/creating-document-store))
+- sample database can be deployed as in any normal instance (more [here](../database/tasks/create-sample-data))
+
+{INFO/}
+
+{WARNING:Important}
+
+- all databases available at `4.live-test.ravendb.net` server are publicly available and can be accessed, managed and copied by any user.
+- periodicaly all databases will be removed
```
ayende: `Periodically`

Documentation/4.0/Raven.Documentation.Pages/client-api/what-is-a-document-store.dotnet.markdown
IdanHaim: The commands is still an issue in the documantation, that's why I left it here
maybe need to change it to operation?

ayende: Yes

Documentation/4.0/Raven.Documentation.Pages/client-api/creating-document-store.markdown
IdanHaim: I will review it and see what can be change.

IdanHaim: I figute out a way, going to talk in this section only about Initialize,
after singleton I will add a Note that just tell the user about dispose,
we don't need to talk about using at all for my opinion the user can see this in the code.

if we still want to talk about using I am gonna remove In a real case scenario it should be done when an application shuts down. line and write something else in the Note.

PR #349, opened 2017-12-06
RDoc-1036 refresh getting-started.markdown, remove installation.markdown

iftahbe:


Documentation/4.0/Raven.Documentation.Pages/start/getting-started.markdown
```diff
@@ -0,0 +1,163 @@
...
+    {  
+        "ServerUrl": "http://127.0.0.1:8080",
+        "Setup.Mode": "None",
+        "DataDir": "RavenData"
+    }  
+ 
+
+If port 8080 is already being used, RavenDB will fail to start and give you an "address in use" error (EADDRINUSE).
+
+You can customize the port and host from the command line by issuing the following command on <strong>Windows</strong> which will run RavenDB on port 8081, avoiding the conflicting port issue:
```
ayende: What is the difference between windows & linux in this case?
Why do this in the command line and not the settings file?

Documentation/4.0/Raven.Documentation.Pages/start/getting-started.markdown
```diff
@@ -0,0 +1,163 @@
...
+
+You are ready to continue to the next section, [Getting to Know RavenDB](getting-to-know).
+
+<hr />
+
+## Installing RavenDB with Docker
+
+If you already have Docker installed, getting started is simple. Just run the following command:
+
+    docker run -d -e UNSECURED_ACCESS_ALLOWED=PrivateNetwork
```
ayende: Nope, we explicitly do no want to have this kind of recommendation in the docs.
We need to ensure that this is running in a secured fashion

Documentation/4.0/Raven.Documentation.Pages/start/getting-started.markdown
iftahbe: The difference is .exe
I show both ways: command line and settings.json. Do you think we should skip this?

ayende: I looked at this multiple times and didn't see it. Maybe just have two commands one after another, so that is clearer? 
And better to show just `settings.json`. @aviviadi will handle the configuration docs.

PR #351, opened 2017-12-06
RDoc-744  Indexes > Map-Reduce indexes

arekpalinski:


Documentation/4.0/Raven.Documentation.Pages/Indexes/map-reduce-indexes.markdown
```diff
@@ -0,0 +1,76 @@
+# Map-Reduce indexes
+
+Map-Reduce indexes allow to perform complex aggregation of data. The first stage, called the map, run over documents and extracts portions of data according to the map function.
```
ppekrol: allow => allows
run => runs
"according to the map function" => "according to the defined mapping function(s)"

Documentation/4.0/Raven.Documentation.Pages/Indexes/map-reduce-indexes.markdown
```diff
@@ -0,0 +1,76 @@
...
+Next, the reduce function is applied to the results of the map phase in order to generate the final results.
```
ppekrol: Maybe: Upon completion of a first phase, reduction is applied to the map results and the final outcome is produced.

Documentation/4.0/Raven.Documentation.Pages/Indexes/map-reduce-indexes.markdown
```diff
@@ -0,0 +1,76 @@
...
+
+The idea behind map-reduce indexing is that aggregation queries using such indexes are very cheap. The aggregation is performed only once and the results are stored inside the index.
+Once new data come into the database or existing documents are modified the map reduce index will keep the aggregation results up-to-date. The aggregations are never calculated during querying.
```
ppekrol: 'map reduce' => map-reduce
The aggregations are never calculated during querying => "The aggregations are never done during querying to avoid expensive calculations that could result in a severe performance degradation"

PR #352, opened 2017-12-06
RDoc-809 : Indexes > Querying > Projections

aviv86:


Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
```diff
@@ -0,0 +1,194 @@
...
+{CODE-TAB:csharp:DocumentQuery projections_1_docquery@Indexes\Querying\Projections.cs /}
+{CODE-TAB:csharp:Index indexes_1@Indexes\Querying\Projections.cs /}
+{CODE-TAB-BLOCK:csharp:RQL}
+from index 'Employees/ByFirstAndLastName'
+select FirstName, LastName
+{CODE-TAB-BLOCK/}
+{CODE-TABS/}
+
+This will issue a query to a database, requesting only `FirstName` and `LastName` from all documents that index entries match query predicate from `Employees/ByFirstAndLastName` index. What does it mean? It means that, if index entry matches our query predicate, then we will try to extract all requested fields from that particular entry and if all requested fields are available in there, then we do not download it from storage. Index `Employees/ByFirstAndLastName` used in above query is not storing any fields so documents will be fetched from storage.
+
```
arekpalinski: Everything here is correct but I think we should first explain what is the benefit of using projections: reduced amount of data to send, you can pick only portion of data you actually need, ability to rename some fields, load additional info etc. (Oren's book: 9.4 Projecting results might be useful)

The server side behavior that we can grab data directly from the index is just a bonus.

arekpalinski: Also the description of projections shouldn't be inside `Select` method panel since it applies to `ProjectInto` as well

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
```diff
@@ -0,0 +1,194 @@
...
+{INFO:Projections and Stored fields}
+If projection function only requires fields that are stored, then document will not be loaded from storage and all data will come from index directly. This can increase query performance (by the cost of disk space used) in many situations when whole document is not needed. You can read more about field storing [here](../../indexes/storing-data-in-index).
+{INFO/}
```
arekpalinski: Put this outside this Select panel

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
```diff
@@ -0,0 +1,194 @@
...
+
+So following above rule, if we create index that stores `FirstName` and `LastName` and request only those fields in query, then **data will come from index directly**.
+
```
arekpalinski: Explain how to store fields. Show such index def

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
```diff
@@ -0,0 +1,194 @@
...
+{CODE-TAB:csharp:Index indexes_4@Indexes\Querying\Projections.cs /}
+{CODE-TAB-BLOCK:csharp:RQL}
+from index 'Orders/ByShippedAtAndCompany' as o
+load o.Company as c
+select {
+	CompanyName: c.Name,
+	ShippedAt: o.ShippedAt
+}
+{CODE-TAB-BLOCK/}
+{CODE-TABS/}
```
arekpalinski: I see `DocumentQuery` tab is missing here and in a few other examples. Isn't it possible to define such queries using `DocumentQuery` syntax?

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
```diff
@@ -0,0 +1,194 @@
...
+{PANEL:ProjectInto}
+
+This extension method retrieves all public fields and properties of the type given in generic and uses them to perform projection to the requested type.
+You can use this method instead of using `Select` together with all fields of the projection class.
+
+### Example
+
+{CODE-TABS}
+{CODE-TAB:csharp:Sync projections_8@ClientApi\Session\Querying\HowToProjectQueryResults.cs /}
+{CODE-TAB:csharp:Async projections_8_async@ClientApi\Session\Querying\HowToProjectQueryResults.cs /}
```
arekpalinski: We decided to show `Async` code as well only in Client API. Let's drop this tab here

arekpalinski: And rename `Sync` tab to `Query` as the others

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
aviv86: ok

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
aviv86: ok

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
aviv86: in some cases it's possible, for instance, if we replace RavenQuery.Load with session.Load in this example then it's possible to define this query using DocumentQuery syntax.

but according to @ayende in his comment here:
http://issues.hibernatingrhinos.com/issue/RavenDB-9652
it's not a valid use case..

so i figured it's better to exclude the DocumentQuery syntax in all the examples that uses 'let' and  in the one with Raw JavaScript Code

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
aviv86: ok

Documentation/4.0/Raven.Documentation.Pages/Indexes/Querying/projections.dotnet.markdown
arekpalinski: - I think the usage of `session.Load` would do the load call to the database. It won't be part of RQL sent to the server. 
- how to use Raw JavaScript Code in projection using `DocumentQuery`?

aviv86: - yes, it does the load call to the database and it's not a part of RQL sent to the server.
- it's not possible to use Raw JavaScript in projection using 'DocumentQuery'.. same as with 'session.Load', it's not a part of RQL sent to the server and it tries to do the call to the database.
according to @ayende  it's not a valid use case:
http://issues.hibernatingrhinos.com/issue/RavenDB-9653

aviv86: actually i just realized now that all my DocumentQuery examples are wrong, since it's not really a projection if i use 'select' - it's not a part of the RQL sent to the server. 
we can probably to do it using SelectFields with the appropriate QueryData, but @ayende says it's too complicated and that it's better to just remove all the DocumentQuery examples.

arekpalinski: ok

PR #377, opened 2017-12-09
RDoc-917: Introduced set-based patch operations

maximburyak:
Updated documents tree to be more clear about the patching operations


maximburyak:
That's ac

PR #378, opened 2017-12-09
RDoc-817: Introduced set-based patch operations

maximburyak:
Updated documents tree to be more clear about the patching operations


Documentation/4.0/Raven.Documentation.Pages/client-api/session/how-to/.docs.json
```diff
@@ -0,0 +1,7 @@
+﻿[
```
ppekrol: Let's remove this. Patching can be in operations only.

PR #390, opened 2017-12-13
 RDoc-745 and fix in  RDoc-844

arekpalinski:


Documentation/4.0/Raven.Documentation.Pages/indexes/stale-indexes.markdown
```diff
@@ -0,0 +1,78 @@
+# Stale Indexes
+
+RavenDB performs indexing in the background threads. The indexes start processing whenever the new data comes in, the existing documents are updated or deleted.
+Running them in the background allows the server to return query results immediately regardless the large number of documents has been just changed.
```
ppekrol: Maybe: regardless if large number of documents has been changed or not

Documentation/4.0/Raven.Documentation.Pages/indexes/stale-indexes.markdown
```diff
@@ -0,0 +1,78 @@
...
+However in that case, the index is stale until it processes them.
```
ppekrol: Maybe: the index is considered stale...

Documentation/4.0/Raven.Documentation.Pages/indexes/stale-indexes.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+{CODE stale1@Indexes\StaleIndexes.cs /}
+
+When `IsStale` is true, that means some `Product` has been just added or changed, and the index didn't have enough time to fully update before the query.
+
+Typically you don't need to worry about it, the index usually updates the records within milliseconds, however there are scenarios where you cannot work with possibly stale data.
+
+## Explicitly Waiting for Non-Stale Results
+
+In order to assure that a query won't return stale results you can use a few approaches. By default RavenDB will wait 15 seconds unless different timeout was specified.
```
ppekrol: I think that we are missing here something that states that the default timeout is for each of the approaches. Now it looks like the 15 second timeout is the default so someone can assume that we are waiting for non stale results by default. I would add explicitly something here to clear that up.

ppekrol: typo: different timeout was specified -> different timeout is specified.

Documentation/4.0/Raven.Documentation.Pages/indexes/stale-indexes.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+The usage of `WaitForNonStaleResults` at the query level is usually reasonable on only rare occasions. 
+Taking advantage of `WaitForNonStaleResults` customization applied to the all queries is very often a symptom of deeper issues in an application model and 
+misunderstanding of the querying concepts in RavenDB. 
+{DANGER/}
+
+
+## Cutoff point
+
+If a query sent to the server specifies it needs to wait for non-stale results then RavenDB sets the cutoff etag for the staleness check.
```
ppekrol: I find this text better than the one (similar) used in how-to-customize-query maybe it should be copied there?

PR #411, opened 2017-12-18
RDoc-902 :  Client API > How to > store dates

aviv86:


arekpalinski:
I think we should mention here that we are ISO 8601 compliant and what are our default storing formats. Please check https://github.com/ravendb/ravendb/blob/v4.0/src/Sparrow/DefaultFormat.cs

Documentation/4.0/Raven.Documentation.Pages/client-api/how-to/store-dates.markdown
```diff
@@ -6,5 +6,19 @@ However if you have people from around the world using the same database, and yo
...
+
+##ISO 8601 Compliance and Default Storing Formats
+
+RavenDB is [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) compliant.   
+
+RavenDB uses the follwing default formats for storing `DateTime`:  
+
+- **"yyyy'-'MM'-'dd'T'HH':'mm':'ss"**
+- **"yyyy'-'MM'-'dd'T'HH':'mm':'ss.fffffff"**
+- **"yyyy'-'MM'-'dd'T'HH':'mm':'ss.fffffff'Z'"**
```
arekpalinski: How can we use 3 formats at the same time to store a single `DateTime` property? We use this one
https://github.com/ravendb/ravendb/blob/v4.0/src/Sparrow/DefaultFormat.cs#L6
don't we?

PR #444, opened 2018-01-04
RDoc-885 Client API > FAQ > Backward compatibility

ppekrol:


Documentation/4.0/Raven.Documentation.Pages/client-api/faq/backward-compatibility.markdown
```diff
@@ -0,0 +1,24 @@
+# Backward compatibility
+
+When it comes to client server compatibility in RavenDB there are two valid rules. Either:
+
+ * The server version is within the same Major and Minor range as the client version
+
+**OR**
+
+ * The server version is greater than the client version
```
ayende: But not between 3.0 and 4.0

ppekrol: done

Documentation/4.0/Raven.Documentation.Pages/client-api/faq/backward-compatibility.markdown
```diff
@@ -6,7 +6,7 @@ When it comes to client server compatibility in RavenDB there are two valid rule
 
 **OR**
 
- * The server version is greater than the client version
+ * The server version is greater than the client version, but the Major version match
```
ayende: We still need to be _explicit_ for users coming from 3.x, that it is not supported.

ppekrol: Added:
`3.x` client will **NOT** work with `4.0` server since client and server Major version do not match

PR #446, opened 2018-01-04
RDoc-1072 documenting test driver

talweiss1982:


Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriver.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Linq;
+using Raven.Client.Documents.Indexes;
+
+namespace RavenDBTestDriver
+{
+    public class RavenDBTestDriver : RavenTestDriver<MyRavenDBLocator>
+    {
+        #region test_driver_2
+        //This allows us to generate a clean database for our test
+        protected override void SetupDatabase(IDocumentStore documentStore)
```
ayende: Why do we need this? Unless I remember badly, we are already doing this for the user.

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriver.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            var doc = new DatabaseRecord(documentStore.Database);
+            try
+            {
+                documentStore.Maintenance.Server.Send(new CreateDatabaseOperation(doc));
+            }
+            catch (Exception e)
+            {
+                //Ignore if the database exists, we could also delete it and create a new one if we need
+                if (e.Message.EndsWith("already exists!") == false)
```
ayende: Let's avoid such error handling.

Documentation/4.0/Raven.Documentation.Pages/start/test-driver.markdown
```diff
@@ -0,0 +1,82 @@
...
+{PANEL:PreInitializingTheStore}
+
+Pre-Initializing the IDocumentStore allows you to mutate the conventions used by the document store.
+### Example
+{CODE test_driver_3@Start\RavenDBTestDriver.cs /}
+
+{PANEL/}
+
+{PANEL:UnitTest}
+Finally we can write down a simple test, note that i'm using [xunit](https://www.nuget.org/packages/xunit/) for my test framework in the below example.
```
ayende: `i'm` -> `I'm`

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriverFull.cs
```diff
@@ -0,0 +1,84 @@
...
+                    var query = session.Query<TestDocument, TestDocumentByName>().Where(x => x.Name == "hello").ToList();
+                    Assert.Single(query);
+                }
+            }
+        }
+    }
+
+    public class MyRavenDBLocator : RavenServerLocator
+    {
+        public override string ServerPath => @"C:\work\RavenDB-4.0.0-rc-40025-windows-x64\Server\Raven.Server.dll";
```
ayende: Add some logic here, so people won't think that there is just a const here.
Something like if search a few paths and return the first that match,etc

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriver.cs
talweiss1982: we only take care of it if paths are provided from what i saw

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriver.cs
talweiss1982: taken from the source code of ravendb

ppekrol: @talweiss1982 I think that it should throw ConcurrencyException (or I might be wrong), we can catch that

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriver.cs
gregolsky: We create a DB for the user already https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.TestDriver/RavenTestDriver.cs#L59
Additionally he's able to provide `DatabaseDumpFilePath` or `DatabaseDumpFileStream` if he wants it imported there.

talweiss1982:
done

Documentation/4.0/Raven.Documentation.Pages/start/test-driver.markdown
```diff
@@ -0,0 +1,75 @@
...
+| **protected virtual void SetupDatabase(IDocumentStore documentStore)** | Allows you to initialize the database. |
+| **protected event EventHandler DriverDisposed;** |An event that is raised when the test driver is been disposed of. |
+| **public void WaitForIndexing(IDocumentStore store, string database = null, TimeSpan? timeout = null)** | Allows you to wait for indexes to become non-stale. |
+| **public void WaitForUserToContinueTheTest(IDocumentStore store)** | Allows you to break the test and launch the studio to examine the state of the database. |
+| **protected virtual void OpenBrowser(string url)** | Allows you to open the browser. |
+| **public virtual void Dispose()** | Allows you to dispose of the server. |
+
+{PANEL/}
+
+{PANEL:PreInitializingTheStore}
```
ppekrol: 'PreInitialize' would be enough?

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriver.cs
```diff
@@ -23,10 +23,6 @@ protected override void SetupDatabase(IDocumentStore documentStore)
             catch (Exception e)
             {
                 //Ignore if the database exists, we could also delete it and create a new one if we need
-                if (e.Message.EndsWith("already exists!") == false)
```
ayende: The point of `SetupDatabase` is to allow you to do things like create a dataset.
There is no need to create db here, and this should be removed.

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Start/RavenDBTestDriverFull.cs
```diff
@@ -61,7 +57,13 @@ public void MyFirstTest()
 
     public class MyRavenDBLocator : RavenServerLocator
     {
-        public override string ServerPath => @"C:\work\RavenDB-4.0.0-rc-40025-windows-x64\Server\Raven.Server.dll";
+        public override string ServerPath
+        {
+            get
+            {
+                return Environment.GetEnvironmentVariable("RavenServerTestPath");
```
ayende: Let's use multiple paths, I want to make it explicit that there can be logic there.
So you will search each of the paths for FileExists

PureKrome:
Thank you!

PR #448, opened 2018-01-04
RDoc-933 : Client API > Commands > Batches > How to send multiple commands using a batch?

aviv86:


Documentation/4.0/Raven.Documentation.Pages/client-api/commands/batches/how-to-send-multiple-commands-using-a-batch.markdown
```diff
@@ -0,0 +1,38 @@
...
+## Example
+
+{PANEL}
+{CODE-TABS}
+{CODE-TAB:csharp:Sync batch_3@ClientApi\Commands\Batches\HowToSendMultipleCommandsUsingBatch.cs /}
+{CODE-TAB:csharp:Async batch_3_async@ClientApi\Commands\Batches\HowToSendMultipleCommandsUsingBatch.cs/}
+{CODE-TABS/}
+{PANEL/}
+
+{NOTE All the commands in the batch will succeed or fail as a **transaction**. Other users will not be able to see any of the changes until the entire batch completes./}
```
arekpalinski: as a transaction - did you mean something like "as they are run in the same transaction"

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Commands/Batches/HowToSendMultipleCommandsUsingBatch.cs
```diff
@@ -0,0 +1,102 @@
...
+        #region batch_2
+        public class BatchOptions
+        {
+            public bool WaitForReplicas { get; set; }
+
+            public int NumberOfReplicasToWaitFor { get; set; }
+
+            public TimeSpan WaitForReplicasTimeout { get; set; }
+
+            public bool Majority { get; set; }
```
arekpalinski: Most properties is self explaining but not this one. Please at least add a comment here explaining what it's used for.

Documentation/4.0/Raven.Documentation.Pages/client-api/commands/batches/how-to-send-multiple-commands-using-a-batch.markdown
ayende: In this case, everything in the batch _is_ a single transaction. It may be merged with other work, but it will work atomically.

PR #459, opened 2018-01-09
RDoc-1085 Document session.Query.Where(exact:true)

ml054:


ppekrol:
Let's do it in the same manner as other articles. Sync / Async / RQL

PR #466, opened 2018-01-10
RDoc-1083

aviv86:


Documentation/4.0/Raven.Documentation.Pages/client-api/session/how-to/ignore-entity-changes.dotnet.markdown
```diff
@@ -0,0 +1,17 @@
+# Session : How to Ignore Entity Changes
+
+In order to mark an entity as one that should be ignored for change tracking purposes, use the `IgnoreChangesFor` method from `Advanced` session operations.  
+The entity will still take part in the session, but will be ignored for `SaveChanges`.
```
arekpalinski: Just to be sure, it means that another `Load` of that entity won't create a call to the server, right? That's the difference from `Evict` method?

aviv86: Yes that's correct

arekpalinski: Maybe let's put such info explicitly here

PR #483, opened 2018-01-15
RDoc-1017 understanding eventual consistency

talweiss1982:


Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/UsersIssues/UnderstandingEventualConsistency.cs
```diff
@@ -0,0 +1,55 @@
...
+            using (var store = new DocumentStore())
+            {
+                using (var session = store.OpenSession())
+                {
+                    var pageSize = 0;
+                    var pageNumber = 0;
+                    var timeToWaitForIndexToUpdateInSec = TimeSpan.FromSeconds(5);
+                    #region userissues_2                    
+                    session.Advanced.WaitForIndexesAfterSaveChanges(timeToWaitForIndexToUpdateInSec);
+                    var results = session.Query<Product>()
```
arekpalinski: What do you think about skipping `session.Query<Product>()... ` completely here? Currently, it might look (especially for newbies)  that calling `WaitForIndexesAfterSaveChanges` before doing the query ensures you won't get stale results.

PR #484, opened 2018-01-15
RDoc-911 Client API > Commands > How To > ...get names of all databas…

ml054:
…es on a server


arekpalinski:
It's the operation. It should be here: https://ravendb.net/docs/article-page/4.0/csharp/client-api/operations/server/get-database-names-operation

PR #530, opened 2018-01-20
Added test-driver CI/CD examples.

PureKrome:
Samples are for:
- AppVeyor
- VSTS


PureKrome:
Some extra notes about this PR.

1. I'm not sure how to use the custom Markdown to do 'code snippets'. Basically, I couldn't figure out where to find the sample files.
2. Right now, it's downloading a fixed/specific raven-server. This should change in the future to some other version and it moves out of RC status. I hate hardcoding things like this - so I'm happy to see what options you guys might consider. Maybe using NuGet is the way to do this. I'm not sure of the NuGet commands to do this.
3. I also don't really like how I'm hardcoding the environmental variables in VSTS but .. you know .. vsts isn't as nice as AV so this might be because I just don't know how to do stuff in VSTS .. compared to my waaay more knowledge of AV.

PureKrome:
Also - here's the sample AV output with the above code, which I used to figure out what to do in AV: https://ci.appveyor.com/project/PureKrome/ravendbtestdriverexampletests

Once the PR is accepted (if it passes acceptance, etc) i'll nuke that repo. Sorry, anyone coming here in the future, via search engines etc :)

ayende:
@PureKrome Here are some of our notes on how to handle this:
http://issues.hibernatingrhinos.com/issue/RavenDB-7787

One thing to note is that you'll download ravendb on _every_ build, as I read the script. 
It's better to check if the file exists and skip the download

PureKrome:
> One thing to note is that you'll download ravendb on every build, as I read the script.
> It's better to check if the file exists and skip the download

@ayende is that in reference to the AV script? Because right now, that file will not exist in _each_ AV build that is fired off unless it is somehow cached. Right now, only the `packages` folder is cached ... but it would be a great idea to cache that file also (if I can figure out how to cache other folders 👍 )

PureKrome:
Ping @ayende - just looking some feedback to my previous reply.

(Trying to get this PR accepted, etc)

ayende:
@gregolsky Can you look at this? I think this can be merged unless you have comments.
You are more familiar with CI env behavior.

Documentation/4.0/Raven.Documentation.Pages/start/test-driver.markdown
```diff
@@ -69,6 +70,131 @@ In the test we get an IDocumentStore to our test database, deploy an index and i
...
+configuration: Release
+os: Visual Studio 2017
+pull_requests:
+  do_not_increment_build_number: true
+environment:
+  RavenServerDirectory: '%temp%\raven-server'
+  RavenServerDownloadDestinationFile: '%temp%\raven-server.zip'
+  RavenServerTestPath: '%RavenServerDirectory%\server\Raven.Server.dll'
+
+init:
```
gregolsky: I don't know neither AppVeyor nor VSTS unfortunately, so I'm gonna ask is the init run only once at job config creation?

Documentation/4.0/Raven.Documentation.Pages/start/test-driver.markdown
```diff
@@ -69,6 +70,131 @@ In the test we get an IDocumentStore to our test database, deploy an index and i
...
+![](images/test-driver/td2.png)
+
+here's the code to quickly copy/paste the script into your VSTS task settings:
+
+```
+Write-Output "Lets see what all our Environmental variables are now defined as:"
+Get-ChildItem Env:
+
+Write-Output "Downloading RavenDb 4.0.0-rc-40025 ..."
+(new-object net.webclient).DownloadFile('https://daily-builds.s3.amazonaws.com/RavenDB-4.0.0-rc-40025-windows-x64.zip',  $env:RavenServerDownloadDestinationFile)
```
gregolsky: You could add simple `Test-Path` checks to see if dir/zip exists already, like:

```powershell
if ((Test-Path $env:RavenServerDownloadDestinationFile) -eq $False) {
   ...download it
}
```

gregolsky:
@PureKrome I don't know AV or VSTS unfortunately. You could check files for existence in those PS scripts. Apart from that it looks good to me 👍

Documentation/4.0/Raven.Documentation.Pages/start/test-driver.markdown
PureKrome: Nope. The `init` step is run once _each time AppVeyor kicks off a job to run_.

With AppVeyor, it's just a common VM with a [list of pre-installed software](https://www.appveyor.com/docs/build-environment/#pre-installed-software) (like .netcore or npm, etc) and then some process which runs through a list of steps, synchronously (called [the build pipeline](https://www.appveyor.com/docs/build-configuration/#build-pipeline) ).

So this means, each time AV kicks off a run, you start out with a vanilla VM (more or less).

You can _offline cache_ any _files_ at the end of the run, if it's quicker to 'copy-restore' those files vs _installing_ or _computing_ (eg. compiling or downloading) stuff during a run. e.g. nuget packages.

So I _could_ cache the zip file being downloaded, but then I would need to download the zip from some azure-blob-backup (which would be very close/local to the VM running) vs download-the-zip from aws. To keep things simple, I chose the later because I'm assuming the time difference wouldn't be too much.

NOTE: I've had a bit of experience with AV and have noticed that the cache-restore process isn't the fastest actually, so I tend to be picky-choosey, when I use it and for what.

Documentation/4.0/Raven.Documentation.Pages/start/test-driver.markdown
PureKrome: Sure could! (refer to my comment, above). But this means I would have to add this step _later_ on (after the cache-restore completes) which then adds a bit more complexity. It would really then come down to a complexity-vs-speed test and get metrics.

I'm also thinking, then times are not going to be a huge difference, either. not like 10 mins vs a few seconds, or something.

PureKrome:
>You could check files for existence in those PS scripts. Apart from that it looks good to me

Kewl! At least this is a starting point for people :) Sure, it can be further refined in the future given some more love-and-time and getting some build-metrics...

Documentation/4.0/Raven.Documentation.Pages/start/test-driver.markdown
gregolsky: OK, for docs let's keep it as-is. People can work it out on their own if they want to.

PureKrome:
Woot!!

PR #534, opened 2018-01-21
RDoc-1106

aviv86:


Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/Indexes/Querying/Searching.cs
```diff
@@ -346,9 +351,9 @@ public Searching()
                 using (var session = store.OpenSession())
                 {
                     #region search_21_0
-                    IList<User> users = session
-                        .Query<User, Users_ByName>()
-                        .Search(x => x.Name, "John")
+                    var users = session
+                        .Query<Users_Search.Result, Users_Search>()
+                        .Search(x => x.Query, "John")
```
arekpalinski: Missing `As<User>()`. Fixed in d03e6ce31f77d05

PR #546, opened 2018-01-24
RDoc-841 changes after review

myarichuk:


Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/SavingChanges.cs
```diff
@@ -33,13 +38,35 @@ public SavingChanges()
 				}
 			}
 
+		    using (var store = new DocumentStore())
+		    {
+		        using (var session = store.OpenAsyncSession())
```
arekpalinski: Just FYI, we typically use `asyncSession` for `IDocumentSession` variables in async examples. I'll fix it

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/SavingChanges.cs
```diff
@@ -74,6 +123,30 @@ public SavingChanges()
...
+		        {
+		            // storing new entity
+		            #region saving_changes_4_async
+
+		            session.Advanced.WaitForReplicationAfterSaveChanges(
+		                timeout: TimeSpan.FromSeconds(30),
+		                throwOnTimeout: false, //default true
+		                replicas:2, //minimum replicas to replicate
+		                majority:false);
```
arekpalinski: Does it make sense to specify `replicas` and `majority` together?

arekpalinski: Nvm, just noticed `majority` has a default `false` value

PR #582, opened 2018-02-09
RDoc-1124 Client API > Operations > compare - exchange

ml054:


arekpalinski:
Maybe let's put some introduction article that will describe what is compare exchange. Mention that it's cluster wide. You can put there the node you are repeating in every article: 

Compare-exchange values could be used on the client side to ensure [unique constraints]

ml054:
Ok I will do that.

arekpalinski:
Please still apply the above comment

PR #586, opened 2018-02-12
RDoc-863 update get-entity-metadata to how-to-get-and-modify-entity-m…

IdanHaim:
…etadata

### DIFF
--- a/Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/HowTo/GetMetadata.cs
+++ b/Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/HowTo/GetMetadata.cs
@@ -1,30 +1,69 @@
-﻿using Raven.Client.Documents;
+﻿using System;
+using System.Collections.Generic;
+using Raven.Client;
+using Raven.Client.Documents;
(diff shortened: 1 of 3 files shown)


Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/HowTo/GetMetadata.cs
```diff
@@ -1,30 +1,69 @@
...
+
+                using (var session = store.OpenSession())
+                {
+                    var user = session.Load<User>("users/1-A");
+                    var metadata = session.Advanced.GetMetadataFor(user);
+
+                    if (metadata.ContainsKey("Permissions"))
+                        metadata["Permissions"] = "ReadAndWrite";
+                    else
+                        metadata.Add("Permissions", "ReadAndWrite");
```
arekpalinski: Wouldn't `metadata["Permissions"] = "ReadAndWrite";` be enough? instead of `if .. else `

IdanHaim: You are right I changed it

PR #603, opened 2018-02-20
RDoc-1135 Studio > Database > Documents > Create New Document

Danielle9897:


Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
```diff
@@ -1,64 +1,171 @@
...
+  The server  
+
+* **Description**:  
+  * When a document ID is _not_ specified, RavenDB server will generate a **globally unique identifier** (GUID) for the stored new document.
+  * Although this is the simplest way to generate a document ID, Guids are _not_ human-friendly when it comes to debugging or troubleshooting 
+    and are less recommended.  
+
+* **When to use**:  
+  Only When you don't care about the exact ID generated and about the ease of troubleshooting your app...
+{PANEL/}
```
arekpalinski: There was a warning that the usage of non-sequential IDs (e.g. Guids) might affect the performance. Would be good to have the following section here as well (3.5):

https://ravendb.net/docs/article-page/3.5/csharp/client-api/document-identifiers/beware-of#non-sequential-identifiers

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
```diff
@@ -1,64 +1,171 @@
...
+  * Use the server-side ID when you don't care about the exact ID that is given to a newly created document.  
+  * Recommended when a large number of documents are needed to be created, such as in bulk insert scenarios,  
+    as this method requires the least amount of work from RavenDB.  
+
+* **Example**:  
+  * From a server running on node 'A':  
+      * Creating the first document with 'users/' => will result with document ID: _'users/0000000000000000001-A'_  
+      * Creating a second document with 'users/' => will result with document ID: _'users/0000000000000000002-A'_  
+  * From a server running on node 'B':  
+      * Creating a third document with 'users/' => will result with document ID: _'users/0000000000000000003-B'_
```
arekpalinski: This suggests the numeric part will be increasing across the cluster what is not true. On server B you might get users/0000000000000000001-B (note 1).

It will be increasing only within _the same node_

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
```diff
@@ -1,64 +1,171 @@
...
+{PANEL/}
+
+{PANEL: Identity}
+
+* **Generated by**:  
+  The server  
+
+* **Description**:  
+  *  Upon document creation, providing a string that ends with a _pipe symbol_ ( | ) for the document ID will cause the server to generate an **identity**.  
+  *  RavenDb will create a simple, always-incrementing value and append it to the ID string provided.
```
arekpalinski: Need to add that it will be unique _across_ the cluster.

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
```diff
@@ -1,64 +1,171 @@
...
+
+{PANEL: HiLo Algorithm}
+
+* **Generated by**:  
+  Both the server and the client  
+
+* **Description**:  
+  * The Hilo algorithm enables generating document IDs on the client.  
+  * The client reserves a range of identifiers from the server.  
+  * The server ensures that this range will be provided only to this client, different clients will receive different ranges.
```
arekpalinski: this client -> single / one client ?

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
```diff
@@ -1,64 +1,171 @@
...
+
+* **Example**:  
+  people/128-A, people/129-B
+{PANEL/}
+
+{PANEL: Document IDs - Limitations}
+
+There are two limitations for document IDs:  
+
+* The maximum identifier length allowed is 512 chars
```
arekpalinski: Where did you get it from? Previous version of article had the correct value (The maximum length allowed is 2025 bytes in UTF8)

Documentation/4.0/Raven.Documentation.Pages/studio/database/documents/create-new-document.markdown
```diff
@@ -0,0 +1,102 @@
...
+
+* From the Studio, a new document can be added to a database to either an existing collection or a new one
+
+* The new document ID can be generated as any one of the following types:  
+  * Server-Side ID  
+  * Semantic ID  
+  * Identity  
+  * GUID  
+
+* An ID generated by the HiLo Algorithm _cannot_ be generated from the Studio, only from the client
```
arekpalinski: only from the client -> only with the usage of RavenDB client

Documentation/4.0/Raven.Documentation.Pages/studio/database/documents/create-new-document.markdown
```diff
@@ -0,0 +1,102 @@
...
+![Figure 1. Create New Document](images/create-new-document-1.png "Create New Document")
+
+* Edit the new document  
+<br/>
+![Figure 2. Edit the New Document](images/create-new-document-2.png "Edit the New Document")
+
+* 1 - **Enter document ID** - see the [examples](../../../studio/database/documents/create-new-document#document-id-types---examples) below  
+  * Any collection name can be entered as a prefix.  
+  * The collection name will be created if it doesn't exist yet.  
+  * if no collection name is specified, i.e. the ID string doesn't end with ( / ) or ( | ),
```
arekpalinski: if -> If

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
Danielle9897: yes - that is true - there was such warning - 
I talked w/ Oren - Oren said to remove this warning from the documentation

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
Danielle9897: Yes - previous version had:  *2025 bytes in UTF8*
On the other hand - the book has:  *512 bytes*
So - I asked Oren which to use - and he answered 512 chars
**So - what is correct ?**

Documentation/4.0/Raven.Documentation.Pages/studio/database/documents/create-new-document.markdown
Danielle9897: fixed

Documentation/4.0/Raven.Documentation.Pages/studio/database/documents/create-new-document.markdown
Danielle9897: fixed

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
arekpalinski: We don't have such a check (512 chars) in raven codebase. 2025 is the voron limitation. @ayende ?

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
ayende: This is correct, because we are now no longer storing the data in the same place as the index.
Case in point, our production system has a database that is about 10 GB in size.
One of the collections is of size  4.22 GB, but the size of the storage indexes to find documents by etag on this collection is 53MB.
The total size of all document ids index is 392.82 MB.
Because we are storing the document's data and its id separately, we no longer really worry about the fragmentation of the ids index. 
This particular database has ids that are about 70 characters each, and uses GUIDs inside them.
It is a good example for how that works.
In this case, we have ~ 3 million records.
Assuming fixed costs of about 90 bytes per record, we can pack the whole thing into about 32,436 pages, or about 253 MB.
In practice, we are probably fragmented heavily here, and we have this in 50,281 pages and just under 400MB.

Either way, this is small enough that we don't really care, and that is for the worst case possible. Entries that are very large and random.

Other databases there have :

* 113,773 entries in 6MB 
* 23,818 entries in 1.4MB (it fits on a floppy!) :-)
* 653,590 entries in 38 MB

So that isn't really important for performance as it used to be

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
ayende: Yes, the limit is 2025 bytes only, no limits on number of characters.

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
Danielle9897: I changed to:

```

Description:

    The Hilo algorithm enables generating document IDs on the client.
    The client reserves a range of identifiers from the server and the server ensures that this range will be provided only to this client.
    Different clients will receive different ranges.
    Each client can then safely generate identifiers within the range it was given, no further coordination with the server is required.
    For a more detailed explanation see HiLo Algorithm

```

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
Danielle9897: Ok, changed to:  ...  *The identifier length limit is 2025 bytes (in UTF8)*

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
Danielle9897: I modified the explanation :)

Documentation/4.0/Raven.Documentation.Pages/server/kb/document-identifier-generation.markdown
Danielle9897: Right - Added !

PR #618, opened 2018-03-03
RavenDB-10580 Suggesting to rename 'Dynamic Node Distribution' to 'Dy…

ml054:
…namic Database Distribution'


ayende:
The images still has dynamic node distribution?

ml054:
Oops, my fault. Will fix that. Thanks for catching this.

PR #624, opened 2018-03-05
RDoc-770 Update setup-aggressive-caching

IdanHaim:


Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/HowTo/SetupAggressiveCaching.cs
```diff
@@ -22,12 +22,29 @@ public SetupAggressiveCaching()
 
                 using (var session = documentStore.OpenSession())
                 {
+                    #region aggressive_cache_global
+
+                    documentStore.AggressivelyCacheFor(TimeSpan.FromMinutes(5));
```
arekpalinski: Put inside `using` to stress out that it applies until you dispose it

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/HowTo/SetupAggressiveCaching.cs
```diff
@@ -22,12 +22,29 @@ public SetupAggressiveCaching()
...
+
+                    documentStore.AggressivelyCache();
```
arekpalinski: please explain in the article or here that it defines the cache duration for 1 day - users don't know what are out defaults

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/HowTo/SetupAggressiveCaching.cs
IdanHaim: This is not suppose to be in using it will stress out only when the store disposed I moved it out of the session to be more clear about that.
like the name of the region global

Documentation/4.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/HowTo/SetupAggressiveCaching.cs
IdanHaim: done

PR #642, opened 2018-03-11
RDoc-1146: add linux example

aviviadi:


Documentation/4.0/Raven.Documentation.Pages/server/storage/directory-structure.markdown
```diff
@@ -47,6 +47,11 @@ The following command will point `Journals` directory of _Northwind_ database to
 C:\RavenDB\Server\RavenData\Databases\Northwind>mklink /J Journals E:\Journals\Northwind
 {CODE-BLOCK/}
 
+#### Linux
+
+{CODE-BLOCK:bash}
+ ln -s ~/RavenDB/Server/RavenData/Databases/Northwind/Journals /mnt/FastDrive/Databases/Northwind/Journals
```
ayende: Will this work for us with symbolic links?

aviviadi: yes, why not?

ayende: `O_DIRECT`, etc? I'm verifying that we don't have to worry about this?

arekpalinski: @aviviadi Also, I used "mount point" in the description regarding Linux - is it the correct / valid term for `ln -s` ?

aviviadi: ln -s : creates soft link to a different location, which can be a on a different mount point.
from : http://www.linfo.org/mount_point.html
"on which an additional filesystem is mounted (i.e., logically attached). " - meaning you can soft link the journals directory to any other directory. The other directory _may_ belong to a different mount point.

aviviadi: @ayende - As for the O_DIRECT, soft links are not reliable. So it is not good for journals.  Only hard links to files.  So it is not good for directories. 
I need to remove/change this section..

ayende: The question is whatever this is a good production config to recommend

ayende: Okay, following up on that. Looks like the docs is good.
We'll need to update the code to handle some edge cases, see: RavenDB-10715

PR #684, opened 2018-04-15
RDoc-564: Lazy Transactions

aviviadi:


Documentation/4.0/Raven.Documentation.Pages/server/storage/lazy-transactions.markdown
```diff
@@ -0,0 +1,12 @@
+# Lazy Transactions 
+##### End Point : /databases/*/admin/lazy-transaction?mode= < true > | < false >
```
arekpalinski: Correct me if I'm wrong but we don't have such endpoint, do we? 
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Server/Documents/Handlers/Admin/TransactionsModeHandler.cs#L12

Aside from that, the article will look better once we provide the description first instead of showing the URL of the relevant endpoint.

arekpalinski: Given that the endpoint allows to set the following modes:
Safe,
Lazy,
Danger

I would suggest  to change the name of the article to `Transaction mode`

Documentation/4.0/Raven.Documentation.Pages/server/storage/lazy-transactions.markdown
```diff
@@ -0,0 +1,12 @@
...
+#### Description
+RavenDB, when Lazy Transaction mode set to `true`, uses temporary mapped files for journals.  This means transactions are written to non persistent memory instead of writing them on the disk (in journal files).
+Only when journal is full (by default 256MB) or when datafile (Raven.voron) is flushed with the journal content - the data is synced from memory onto the physical drive. 
+Therefor, if RavenDB is shutdown abnoramally during Lazy Transaction mode, the last unsynced journal transactions will be lost.  The benfite is higher writing speed to journals on slow IO machines.  At anycase when RavenDB starts up again and perform journal recovery, the previous lazy transaction mode session should not affect / corrupt the already synced data.
+Setting mode to `false` will write the data in memory to the disk (in the journal file).
```
arekpalinski: The description should discuss the differences between the modes that we have: Safe, Lazy, Danger. Let's describe each of them here and underline that Safe is out default behavior.

Documentation/4.0/Raven.Documentation.Pages/server/storage/lazy-transactions.markdown
```diff
@@ -0,0 +1,12 @@
+# Lazy Transactions 
+##### End Point : /databases/*/admin/lazy-transaction?mode= < true > | < false >
+#### Description
+RavenDB, when Lazy Transaction mode set to `true`, uses temporary mapped files for journals.  This means transactions are written to non persistent memory instead of writing them on the disk (in journal files).
+Only when journal is full (by default 256MB) or when datafile (Raven.voron) is flushed with the journal content - the data is synced from memory onto the physical drive. 
+Therefor, if RavenDB is shutdown abnoramally during Lazy Transaction mode, the last unsynced journal transactions will be lost.  The benfite is higher writing speed to journals on slow IO machines.  At anycase when RavenDB starts up again and perform journal recovery, the previous lazy transaction mode session should not affect / corrupt the already synced data.
```
arekpalinski: Therefor -> Therefore

arekpalinski: abnoramally -> abnormally

Documentation/4.0/Raven.Documentation.Pages/server/storage/lazy-transactions.markdown
```diff
@@ -0,0 +1,12 @@
+# Lazy Transactions 
+##### End Point : /databases/*/admin/lazy-transaction?mode= < true > | < false >
+#### Description
+RavenDB, when Lazy Transaction mode set to `true`, uses temporary mapped files for journals.  This means transactions are written to non persistent memory instead of writing them on the disk (in journal files).
+Only when journal is full (by default 256MB) or when datafile (Raven.voron) is flushed with the journal content - the data is synced from memory onto the physical drive.
```
arekpalinski: WRT "by default 256MB" -> we don't provide a way to change it, we have 256 for 64 bits and 32 for 32 bits. See:
https://github.com/ravendb/ravendb/blob/4e7619283e5fbe3bdfaf9b4916c83633fba5d2ab/src/Voron/StorageEnvironmentOptions.cs#L191

Documentation/4.0/Raven.Documentation.Pages/server/storage/lazy-transactions.markdown
arekpalinski: WRT "should not" - let's avoid such wording in the docs,  aren't we should that it works this way?

Documentation/4.0/Raven.Documentation.Pages/server/storage/lazy-transactions.markdown
```diff
@@ -0,0 +1,12 @@
...
+##### End Point : /databases/*/admin/lazy-transaction?mode= < true > | < false >
+#### Description
+RavenDB, when Lazy Transaction mode set to `true`, uses temporary mapped files for journals.  This means transactions are written to non persistent memory instead of writing them on the disk (in journal files).
+Only when journal is full (by default 256MB) or when datafile (Raven.voron) is flushed with the journal content - the data is synced from memory onto the physical drive. 
+Therefor, if RavenDB is shutdown abnoramally during Lazy Transaction mode, the last unsynced journal transactions will be lost.  The benfite is higher writing speed to journals on slow IO machines.  At anycase when RavenDB starts up again and perform journal recovery, the previous lazy transaction mode session should not affect / corrupt the already synced data.
+Setting mode to `false` will write the data in memory to the disk (in the journal file). 
+
+#####Example
+```
+http://127.0.0.1:8080/databases/myDB/admin/lazy-transaction?mode=true
```
arekpalinski: Please use CURL in the example

arekpalinski: BTW Is it intentional that we don't expose a configuration option for setting the transaction mode?
https://ravendb.net/docs/article-page/4.0/csharp/server/configuration/storage-configuration

arekpalinski: The following option needs to be mentioned and linked here:

https://ravendb.net/docs/article-page/4.0/csharp/server/configuration/storage-configuration#storage.transactionsmodedurationinmin

Also its description "How long transaction mode (Danger/Lazy) .... " needs to link to this article. 

It's _very_ counter intuitive that setting the Danger/Lazy mode will expire after 24 hours by default. We need to write about that here to avoid users' confusion

Documentation/4.0/Raven.Documentation.Pages/server/storage/lazy-transactions.markdown
arekpalinski: Also, we need to state explicitly that setting the transaction mode also affects the indexes

PR #693, opened 2018-04-16
RDoc-1026 Server > Administration > Index administration

arekpalinski:


Documentation/4.0/Raven.Documentation.Pages/server/administration/index-administration.markdown
```diff
@@ -0,0 +1,79 @@
+﻿# Administration : Index administration
+
+RavenDB indexes can be administrated easily by the user with the Studio or via maintenance operations in the Client API.
```
ayende: Some of these are node local operations, should probably be mentioned.
Some of these are cluster global, same.

arekpalinski: Added `Operation scope` to each section - Local node or Cluster

PR #714, opened 2018-05-11
fixed incorrect indentation

csharpsi:
The previous commit was using spaces instead of tabs, which I think is what caused the incorrect indentation in the website


ppekrol:
Thanks again!

PR #772, opened 2018-08-21
 RDoc-1320 refresh setup wizard and setup examples pages

iftahbe:


arekpalinski:
I cannot find where those two enhancements are documented (mentioned in RDoc-1320):
- ability to change node tag
- the fact that we generate wildcard certs since 4.1

arekpalinski:
It was probably added in the first commit, right? If that's the case then in the future please apply the changes to a copied article in a separate commit

iftahbe:
I didn't document it explicitly.
The ability to change a node tag is self-explanatory. 
I did remove a part of the 4.0 article which says to setup manually in case you want another node tag.

Also, I didn't find a suitable place to document the fact that we generate wildcard certs since 4.1
It's a feature which is transparent to the user.
I think it's better if we write it in the "what's new" page.

PR #773, opened 2018-08-21
 RDoc-1321 Document get revision at a specific date

fitzchak:


Documentation/4.1/Raven.Documentation.Pages/server/extensions/revisions.markdown
```diff
@@ -49,7 +49,7 @@ Now, let's modify the original document. This would create another revision:
 As you can see, we have a full audit record of all the changes that were made to the document.
 
 You can access the revisions of a specific document by the document's ID ("users/1").
-Or you can access a specific revision by its change vector.
+Or you can access a specific revision by its change vector or by a specific date.
```
arekpalinski: According to this comment:
https://github.com/ravendb/ravendb/pull/7410/files#diff-ed8582902fbc0f1ce5a42fb55f9d2bb7R39

it returns the revision "before _OR_ at the specified date". Please document more precisely what does it actually returns

Documentation/4.1/Samples/csharp/Raven.Documentation.Samples/Server/Revisions.cs
```diff
@@ -76,6 +76,11 @@ public async Task Sample()
                         .Advanced
                         .Revisions
                         .GetAsync<User>(revisionsMetadata[0].GetString(Constants.Documents.Metadata.ChangeVector));
+
+                    User revisonAtYearAgo = await session
+                        .Advanced
+                        .Revisions
+                        .GetAsync<User>("users/1", DateTime.Now.AddYears(-1));
```
arekpalinski: Might be good idea to add a comment here which will state explicitly that it will return the revision that happened before or at the specified date

fitzchak: I think it is better to make this simple: this would return the revision at a specific date. It is implicitly that if there is a revision at this date we would return it, otherwise we would look for the last revision before.

Documentation/4.1/Raven.Documentation.Pages/server/extensions/revisions.markdown
fitzchak: This would return the revision at a specific date.

ppekrol: The issue here is that the text is not precise. We need to stress out in docs what this method does exactly.

PR #852, opened 2018-11-07
Studio > Database > Indexes > Create new map-reduce index

Danielle9897:


Documentation/4.0/Raven.Documentation.Pages/studio/database/Indexes/create-map-reduce-index.markdown
```diff
@@ -0,0 +1,209 @@
...
+* Artificial documents are updated whenever the index completes indexing a batch of documents.  
+{NOTE/}
+
+{NOTE: }
+#### Artificial Documents Usage
+
+* You can set up indexes on top of the Artificial Documents collection, including additional MapReduce indexes,  
+  giving you the option to create recursive map-reduce operations.  
+
+* You can set up an [External Replication](../../../studio/database/tasks/ongoing-tasks/external-replication-task)
```
arekpalinski: The replication will filter them out. Did you mean Raven ETL, maybe?

Danielle9897: Yes - you are right - I will change -
This means I will need to change this in the Book as well...... :)

Documentation/4.0/Raven.Documentation.Pages/studio/database/Indexes/create-map-reduce-index.markdown
```diff
@@ -0,0 +1,210 @@
...
+
+{NOTE: }
+#### Artificial Documents Usage
+
+* You can set up indexes on top of the Artificial Documents collection, including additional MapReduce indexes,  
+  giving you the option to create recursive map-reduce operations.  
+
+* You can set up a [RavenDB ETL Task](../../../studio/database/tasks/ongoing-tasks/ravendb-etl-task) 
+  on the Artificial Documents collection to a dedicated database, 
+  (on the same cluster or on a separate one) for further processing,
```
arekpalinski: WRT "on the same cluster or on a separate one" - it's isn't good idea to setup ETL on artificial docs (or any docs) to a node inside the cluster. The result would be that ETL sends docs to a node which in turn are replicated back due to replication between nodes inside the cluster. It would create a cycle.

Danielle9897: Fixed. Thanks.

PR #880, opened 2018-11-29
[node.js] querying basics + fixes

gregolsky:


gregolsky:
need to fix the links, will reopen later

PR #909, opened 2019-02-20
Mistake in docs

georgiosd:


ml054:
I think sync example is correct, but the mistake is in RQL section:

https://ravendb.net/docs/article-page/4.1/Csharp/client-api/operations/compare-exchange/overview

image

so object id is: users/1, with field email: user@example.com

Query we query, we should use field email, not id:

image

georgiosd:
@ml054 are you sure? It doesn't make sense to me.

The user model entity is identified by its id e.g. `users/1` and we want to create a unique association to its email `foo@bar.com`

So in that way it makes sense that the CompareExchange doc will be `email/foo@bar.com` and store the value of the user id i.e. `users/1`

cc @ayende

ml054:
@georgiosd I'm sorry you are right. 

RavenQuery.CmpXchg<string>("emails/ayende@ayende.com") returns "users/1", so we should compare id with returned value.

ayende:
@ml054 I think that @georgiosd is correct, the result of `RavenQuery.CmpXchg<string>("emails/ayende@ayende.com")` is going to be the `users/1`, so we need to compare it to the id.
The RQL is correct, the code isn't

arekpalinski:
Also confirmed by @karmeli87

georgiosd:
@ml054 no problem, it's why we're all here :)

PR #923, opened 2019-03-08
RDoc-1476 Explain that Query Optimizer is only searching for relevant…

arekpalinski:
… Auto-Indexes


arekpalinski:
@Danielle9897 Fixed in https://github.com/ravendb/docs/pull/923/commits/a53d91505299aa2e7133fb8b8eecfaf96529f7e7

PR #930, opened 2019-03-10
RDoc-1333: lldb linux debugging (KB article)

aviviadi:


ppekrol:
@reebhub please review and reorganize/beautify this article

aviviadi:
Due to the progress dotnet made, we might want to change the technical details in here

aviviadi:
I've decided not to use this as lldb and dotnet made their significant progress. all one needs is to install lldb (5.0) and use it against our RavenDB published instance.

PR #1003, opened 2019-03-29
RDoc-1477 Sorting with dynamic group by queries

arekpalinski:


Documentation/4.0/Raven.Documentation.Pages/client-api/session/querying/how-to-perform-group-by-query.dotnet.markdown
```diff
@@ -141,6 +141,39 @@ In order to use the above extension methods you need to add the following **usin
 
 {PANEL/}
 
+{PANEL: Sorting}
```
ppekrol: It would be good to link from https://ravendb.net/docs/article-page/4.2/csharp/indexes/querying/sorting to this section.

We can add a small new chapter there linking here and an entry in related articles.

PR #1050, opened 2019-07-03
RDoc-1585

morhilai:


Documentation/4.2/Raven.Documentation.Pages/cloud/cloud-backup.markdown
```diff
@@ -1,4 +1,4 @@
-# RavenDB on the Cloud: Backup
+#RavenDB on the Cloud: Backup
```
arekpalinski: Please change `RavenDB on the Cloud` to `Cloud` (so it will be consistent with menu option)

arekpalinski: In all headers

morhilai: Fixed

PR #1061, opened 2019-07-28
R doc 1604 dynamic object

morhilai:


Documentation/4.2/Samples/csharp/Raven.Documentation.Samples/ClientApi/Configuration/Conventions.cs
```diff
@@ -75,7 +75,7 @@ public void Examples()
                     PropertyNameConverter = mi => FirstCharToLower(mi.Name)
                     #endregion
                     ,
-                    #region TopologyCacheLocation
+                    #region AddIdFieldToDynamicObjects
```
maximburyak: I think that you should keep the TopologyCacheLocation region

morhilai: It's still there, lines 69-71. I think a typo I made and then fixed shows up in the commit.

maximburyak: right

maximburyak:
LGTM

PR #1070, opened 2019-09-02
R doc 1624 aggressive

morhilai:
AggressiveCacheMode documented for .NET


Documentation/4.1/Raven.Documentation.Pages/client-api/how-to/setup-aggressive-caching.dotnet.markdown
```diff
@@ -42,6 +46,16 @@ which is equivalent to:
 
 {CODE aggressive_cache_for_one_day_2@ClientApi\HowTo\SetupAggressiveCaching.cs /}
 
+###Disable Change Tracking
+
+You can choose not to update the cache in response to notifications from the server by changing the `AggressiveCacheMode`.
+
+The modes are:  
+* `TrackChanges` - the default value, which allows the cache to update in response to change notifications from the server.
```
arekpalinski: To be more exact, once we get a notification about any document change (put or delete) or index change (batch completed or index removal) then we invalidate items in cache. In result we do the call to the server anyway, regardless AggressiveCaching is set.

arekpalinski: Maybe worth to mention that tracking is done with the usage of Changes API?

Documentation/4.1/Raven.Documentation.Pages/client-api/how-to/setup-aggressive-caching.dotnet.markdown
```diff
@@ -42,6 +46,16 @@ which is equivalent to:
 
 {CODE aggressive_cache_for_one_day_2@ClientApi\HowTo\SetupAggressiveCaching.cs /}
 
+###Disable Change Tracking
+
+You can choose not to update the cache in response to notifications from the server by changing the `AggressiveCacheMode`.
+
+The modes are:  
+* `TrackChanges` - the default value, which allows the cache to update in response to change notifications from the server.  
+* `DoNotTrackChanges` - do not update the cache for duration of aggressive caching.
```
arekpalinski: What do you think about explicitly stating here that this is about not using Changes API at all then, so only relying on provided cache duration then?

Documentation/4.1/Raven.Documentation.Pages/client-api/how-to/setup-aggressive-caching.dotnet.markdown
morhilai: Fixed

Documentation/4.1/Raven.Documentation.Pages/client-api/how-to/setup-aggressive-caching.dotnet.markdown
morhilai: Done

PR #1082, opened 2019-09-24
R doc 1638 shutdown confirm

morhilai:


Documentation/4.2/Raven.Documentation.Pages/server/administration/.docs.json
```diff
@@ -7,7 +7,7 @@
   {
     "Path": "cli.markdown",
     "Name": "CLI (Command Line Interface)",
-    "DiscussionId": "9a757074-8bee-4e4c-bab9-ec7217e881f3",
+    "DiscussionId": "b30082a5-b80f-419a-b27b-a8ef4504dbf5",
```
arekpalinski: why did you change it? we must not change `DiscussionId` for existing articles because Disqus is using it for tracking comments

PR #1102, opened 2019-11-10
RDoc-1671: New page for the operation to remove obsolete entries from…

morhilai:
… the database change vector.

- Copied Change Vector overview page, changed nothing besides adding this text linking to the new page:

"If an instance is no longer being used, its entry can be removed from the change vector to save space using the [UpdateUnusedDatabasesOperation]()"


Documentation/4.2/Raven.Documentation.Pages/client-api/operations/maintenance/clean-change-vector.markdown
```diff
@@ -0,0 +1,66 @@
+﻿# Clean Change Vector
+
+---
+
+{NOTE: }
+
+* A database's [change vector](../../../server/clustering/replication/change-vector) contains entries from each instance of the database 
+in the database group. However, even when an instance goes down its entry will remain in the database change vector. These entries can
```
karmeli87: I would maybe replace "instance goes down" to "instance doesn't exist anymore, because of removal, dynamic replacement or simply importing/restoring to new node"

morhilai: Done

PR #1110, opened 2019-11-28
RDoc-1535: added SelectFields to ClientAPI>HowToProject and Indexes>Projections

morhilai:
-also fixed a link I noticed was broken in indexes/querying/projections.java.markdown

### DIFF
--- a/Documentation/4.0/Raven.Documentation.Pages/client-api/session/querying/how-to-project-query-results.dotnet.markdown
+++ b/Documentation/4.0/Raven.Documentation.Pages/client-api/session/querying/how-to-project-query-results.dotnet.markdown
@@ -137,18 +137,40 @@ select {
 
 {PANEL:SelectFields}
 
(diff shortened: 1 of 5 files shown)


Documentation/4.0/Raven.Documentation.Pages/client-api/session/querying/how-to-project-query-results.dotnet.markdown
```diff
@@ -134,6 +135,25 @@ select {
 
 {PANEL/}
 
+{PANEL:SelectFields}
+
+The `SelectFields` method can only be used with the [Document Query](../../../client-api/session/querying/document-query/what-is-document-query).  
+The fields of the projection are specified as a `string` array of field names or paths. It takes the type of the projection as
```
maximburyak: There are actually three different overloads of that function, one of them uses reflection also to choose the right fields.
We'll need to add this remark and appropriate examples

PR #1131, opened 2020-03-05
RDoc-1697 java index query code samples

morhilai:
In both Java and Csharp, there were problems in the code samples for index definitions:
- Java indexes had nested `Result` classes that were not necessary (they were probably copied from the Csharp code where they are necessary)
- Csharp indexes did not use the Result class in the map function

### DIFF
--- a/Documentation/4.1/Samples/java/src/test/java/net/ravendb/Indexes/Querying/Highlights.java
+++ b/Documentation/4.1/Samples/java/src/test/java/net/ravendb/Indexes/Querying/Highlights.java
@@ -18,7 +18,7 @@ public class Highlights {
 
     //region highlights_1
     public static class BlogPosts_ByContent extends AbstractIndexCreationTask {
-        public static class Result {
+        public static class Projection {
             private String content;
 
             public String getContent() {
@@ -41,27 +41,6 @@ public class Highlights {
 
     //region highlights_7
     public static class BlogPosts_ByCategory_Content extends AbstractIndexCreationTask {
-        public static class Result {
-            private String category;
-            private String content;
-
-            public String getCategory() {
-                return category;
-            }
-
-            public void setCategory(String category) {
-                this.category = category;
-            }
-
-            public String getContent() {
-                return content;
-            }
-
-            public void setContent(String content) {
-                this.content = content;
-            }
-        }
-
         public BlogPosts_ByCategory_Content() {
             map = "docs.Posts.Select(post => new { post.category, post.content })";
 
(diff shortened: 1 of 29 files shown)


Documentation/4.1/Samples/java/src/test/java/net/ravendb/Indexes/MapReduceIndexes.java
```diff
@@ -235,26 +106,26 @@ public MapReduceIndexes() {
         try (IDocumentStore store = new DocumentStore()) {
             try (IDocumentSession session = store.openSession()) {
                 //region map_reduce_0_1
-                List<Products_ByCategory.Result> results = session
-                    .query(Products_ByCategory.Result.class, Products_ByCategory.class)
+                List<Product> results = session
```
ml054: it is map reduce so I think it isn't correct.

Documentation/4.1/Samples/java/src/test/java/net/ravendb/Indexes/MapReduceIndexes.java
```diff
@@ -235,26 +106,26 @@ public MapReduceIndexes() {
...
                     .whereEquals("Category", "Seafood")
                     .toList();
                 //endregion
             }
 
             try (IDocumentSession session = store.openSession()) {
                 //region map_reduce_1_1
-                List<Products_Average_ByCategory.Result> results = session
-                    .query(Products_Average_ByCategory.Result.class, Products_Average_ByCategory.class)
+                List<Product> results = session
```
ml054: same here

Documentation/4.1/Samples/java/src/test/java/net/ravendb/Indexes/MapReduceIndexes.java
```diff
@@ -235,26 +106,26 @@ public MapReduceIndexes() {
...
                     .whereEquals("Category", "Seafood")
                     .toList();
                 //endregion
             }
 
             try (IDocumentSession session = store.openSession()) {
                 //region map_reduce_2_1
-                List<Product_Sales.Result> results = session
-                    .query(Product_Sales.Result.class, Product_Sales.class)
+                List<Product> results = session
```
ml054: same here

Documentation/4.1/Samples/java/src/test/java/net/ravendb/Indexes/Querying/Highlights.java
```diff
@@ -127,9 +106,9 @@ public Highlights() {
                 highlightingOptions.setPreTags(new String[] { "**" });
                 highlightingOptions.setPostTags(new String[] { "**" });
                 highlightingOptions.setGroupKey("category");
-                List<BlogPosts_ByCategory_Content.Result> results = session
+                List<BlogPost> results = session
```
ml054, line 109: I not sure it is correct. Can we skype about that?

morhilai: Fixed

PR #1132, opened 2020-03-08
RDoc-1715-spatial-queries

morhilai:
- Added explanation that WKT shapes used in spatial queries must have the vertices listed in counter-clockwise order


morhilai:
I used this picture for the explanation:

PolygonClock

PR #1151, opened 2020-05-06
R doc 1703 time series

reebhub:
TheTime-Series API is ready for an initial review.
(This is not for an overall review, just for the API - other articles are not committed yet)


reebhub:
better version ready

PR #1178, opened 2020-07-26
RDoc-1739-broken-link

morhilai:
- 5.0 version of `what-is-rql.markdown` was missing sections for keywords `with` and `match`
- 4.2 version of the page was missing a few different things, including a Query Optimizer section, so it is completely replaced with text of the 5.0 version

The result is I haven't added any new text, I just combined parts of the 4.2 and 5.0 versions of `what-is-rql`, because both had stuff missing, and now they are complete and both are identical.

- So now the 4.2 version of `indexes-overview.markdown` no longer has that broken link to the Query Optimizer section.

### DIFF
--- a/Documentation/4.2/Raven.Documentation.Pages/indexes/querying/what-is-rql.markdown
+++ b/Documentation/4.2/Raven.Documentation.Pages/indexes/querying/what-is-rql.markdown
@@ -68,11 +68,10 @@ time and return results at the same speed regardless of the size of the data.
 You can read more about indexes [here](../indexing-basics).   
 
 
-{NOTE: Exceptions: Time-Series and Counters}
-Queries in RavenDB will always use an index, with the exception of [distributed time-series]() 
-and [distributed counters]().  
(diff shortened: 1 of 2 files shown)


Documentation/4.2/Raven.Documentation.Pages/indexes/querying/what-is-rql.markdown
```diff
@@ -1,139 +1,148 @@
...
+RavenDB doesn't include support for full scans. If an index cannot be found for a query, 
+the query optimizer will **create a new index for the query**.  
+
+RavenDB queries will use the index they have created or found, to minimize response 
+time and return results at the same speed regardless of the size of the data.  
+
+You can read more about indexes [here](../indexing-basics).   
+
+
+{NOTE: Exceptions: Time-Series and Counters}
```
arekpalinski, line 71: This it v4.2 file but it says about Time Series available only in 5.0

PR #1187, opened 2020-08-11
RDoc 1757 counters static indexes

morhilai:


Documentation/5.0/Raven.Documentation.Pages/indexes/indexing-counters.dotnet.markdown
```diff
@@ -0,0 +1,72 @@
+# Indexes: Indexing Counters
```
ppekrol, line 1: This should be moved to 'Document Extensions -> Counters' and this article here should just link to the new one. Thoughts? cc @reebhub

reebhub: right

morhilai: I don't like the idea of having an empty page that just links to another, and the previous version of this page is outdated. I suggest just copying the content to Document Extensions -> Counters. That way they see the same info whether they look for it in the indexes section or the doc extensions section. What do you think? @ppekrol

Documentation/5.0/Raven.Documentation.Pages/document-extensions/counters/.docs.json
```diff
@@ -89,5 +89,20 @@
             "Key": "client-api/session/counters/counters-in-clusters"
         }
     ]
+  },
+  {
+    "Path": "indexing-counters.markdown",
```
ppekrol, line 94: Please match what we have in 'TimeSeries'. It should be just `indexing.markdown` right?

morhilai: You're right, it's redundant. Fixed.

PR #1192, opened 2020-08-17
RDoc-1763

reebhub:
session-context load balancing documentation (API & Studio)

PR #1205, opened 2020-08-23
R doc 1773 custom id separator

morhilai:
Documented `IdentityPartsSeparator` in document store conventions


ppekrol:
Do we have a section where we are documenting Client Configuration? If yes, we need to update it too with that.

morhilai:
@ppekrol If you mean the studio pages for Client Configuration, @reebhub is updating those in https://github.com/ravendb/docs/pull/1192 and will include IdentityPartsSeparator. Other than that, I'm not sure where else it can be mentioned.

PR #1210, opened 2020-08-25
R doc 1782 restart embedded server

morhilai:
Documented EmbeddedServer.RestartServerAsync(), GetServerProcessIdAsync(), ServerProcessExited

From https://github.com/ravendb/ravendb/pull/10190/

### DIFF
--- a/Documentation/4.2/Raven.Documentation.Pages/server/Embedded.dotnet.markdown
+++ b/Documentation/4.2/Raven.Documentation.Pages/server/Embedded.dotnet.markdown
@@ -161,10 +161,7 @@ Use `<embedded server>.ServerProcessExited` to observe when the server has crash
 event EventHandler<ServerProcessExitedEventArgs>? ServerProcessExited;
 {CODE-BLOCK/}
 
-Uses arguments of type `ServerProcessExitedEventArgs`, which inherits from `System.EventArgs` 
-and has no parameters - leave empty:  
-
-{CODE:csharp server_process_exited@Server\Embedded.cs /}
+Event data is of type `ServerProcessExitedEventArgs`.
 
 ---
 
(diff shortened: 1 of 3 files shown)


Documentation/4.2/Raven.Documentation.Pages/server/Embedded.dotnet.markdown
```diff
@@ -118,10 +139,34 @@ In case you're not interested in installing the .Net run-time environment on you
...
+
+#### ServerProcessExited Event
+
+Use `<embedded server>.ServerProcessExited` to observe when the server has crashed or exited.  
+
+{CODE-BLOCK:csharp}
+event EventHandler<ServerProcessExitedEventArgs>? ServerProcessExited;
+{CODE-BLOCK/}
+
+Uses arguments of type `ServerProcessExitedEventArgs`, which inherits from `System.EventArgs`
```
ppekrol, line 164: Describe here what they contain. What values, properties. Remove 'and has no parameters - leave empty' not sure what does it mean. Users typically do not create those. Just subscribe to the events.

morhilai: I just meant that [ServerProcessExitedEventArgs](https://github.com/ravendb/ravendb/blob/b8163cb42d94f9a1799fee27454cd77e9bb4e30c/src/Raven.Embedded/ServerProcessExitedEventArgs.cs) is a totally empty class.

Should I it be:
>Uses arguments of type `ServerProcessExitedEventArgs`, which inherits from `System.EventArgs` and has no properties.  
Or just:
>>Uses arguments of type `ServerProcessExitedEventArgs`, which inherits from `System.EventArgs`.
Or not mention it at all?
Just above this, lines 156-162 might be self explanatory on their own.

ppekrol: Just write that the event data is of type 'ServerProcessExitedEventArgs' it is more than enough.

morhilai: Done

PR #1211, opened 2020-08-25
RDoc-1783-docStore-interface-events

morhilai:
Documented that events can be accessed through the interface IDocumentStore in addition to DocumentStore instances


ppekrol:
No need to change anything. Events are already documented. They were not exposed via interface.

PR #1271, opened 2020-11-08
RDoc-1838-scaling documented scaling in overview-and-syntax.markdown

morhilai:


Documentation/5.1/Raven.Documentation.Pages/document-extensions/timeseries/querying/overview-and-syntax.markdown
```diff
@@ -0,0 +1,228 @@
...
+
+{PANEL/}
+
+{PANEL: Scaling Query Results}
+
+Time series query results can be **scaled**, multiplied by some number. This doesn't 
+change the values themselves, only the output of the query. This can serve as a stage 
+in a data processing pipeline, or just for the purposes of displaying the data in a 
+more undertandable format.  
+
```
aviv86, line 88: Scaling is also useful when you choose to lower the precision of your entries values,
in order to [achieve a better compression](https://ravendb.net/docs/article-page/5.0/csharp/document-extensions/timeseries/design#time-series-entries)
e.g. instead of storing values : `0.00005, 0.00006, 0.00003` 
we can store them as : `0.5, 0.6, 0.3` (multiplied by 10_000) and get a better compression (less segments)

In such cases, when we query the timeseries data we can use `Scale` to show the "real" values (original precision)
e.g. `Scale(1/10_000)`

aviv86: yes, looks good 
but scale by `1/(10^6)` to restore original value

aviv86: yes, looks good
but scale by `1/(10^6)`  to restore original value

morhilai: It's a tough choice. I chose 0.000001 in the end.

aviv86 (approved):

PR #1295, opened 2020-12-30
R doc 1852 encryption js console

morhilai:

PR #1296, opened 2020-12-30
R docs 922 923 grammar

morhilai:


Documentation/4.0/Raven.Documentation.Pages/server/security/encryption/secret-key-management.markdown
```diff
@@ -83,6 +91,41 @@ Then, run the following put-key command for **every** encrypted database. Supply
...
+the master key is read only, and can be accessed with this script:  
+
+{CODE-BLOCK:javascript }
+return database.MasterKey
+{CODE-BLOCK/}
+
+A master key can be modified with the script:  
+
+{CODE-BLOCK:javascript }
+server.ServerStore.PutSecretKey(string base64, string name, bool overwrite)
```
ppekrol, line 118: I think we need to remove 'string' and 'bool' from here. This is not a proper JS. Types are defined below anyway.

morhilai: This comment applies to a commit that belongs to a different issue - RDoc-1852 - which was not supposed to be in this PR 
😅. This pull request has now been rebased. The above problem will be addressed in https://github.com/ravendb/docs/pull/1295

PR #1312, opened 2021-03-01
Use static EntityToBlittable for object conversion

gregorylyons:
Code example does not work, refactor in client has moved `session.Advanced.EntityToBlittable` to a static class `EntityToBlittable`

Change was in 2017, for 4.0.
https://github.com/ravendb/ravendb/commit/63110d091b5609ff3e56abba08c4a883829250da

I have not checked the rest of the docs for the same issue (yet!).


gregorylyons:
Strike that, should have hit compile first >_> Will chase down how it works and raise a better PR.

PR #1370, opened 2021-08-20
Adding some clarity to the consumption of subscriptions

georgiosd:
Hey guys - made some changes to the docs that I thought add clarity - but I understand it's very subjective so let me know what you think.


georgiosd:
I see that the build failed but I don't think that was me 🤣

ayende:
In terms of content, looks okay. In terms of structure , how it works, @reebhub will check

ppekrol:
Fixed

PR #1381, opened 2021-10-28
RDoc -1944 IsLoaded returns true when document doesn't exist

serezhub:
Added sentence to clarify why `IsLoaded` returns `true` in C#, Java, and JS from versions 3.0 till 5.2

### DIFF
--- a/Documentation/3.0/Raven.Documentation.Pages/client-api/session/loading-entities.dotnet.markdown
+++ b/Documentation/3.0/Raven.Documentation.Pages/client-api/session/loading-entities.dotnet.markdown
@@ -183,7 +183,9 @@ Entities can be streamed from server using one of the following `Stream` methods
 
 {PANEL:IsLoaded}
 
-To check if entity is attached to session, e.g. has been loaded previously, use `IsLoaded` method from `Advanced` session operations.
+To check if an entity is attached to a session, e.g. it has been loaded previously, use the `IsLoaded` method from the `Advanced` session operations.  
+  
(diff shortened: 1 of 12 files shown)


haludi (approved):

Documentation/3.0/Raven.Documentation.Pages/client-api/session/loading-entities.dotnet.markdown
```diff
@@ -183,7 +183,9 @@ Entities can be streamed from server using one of the following `Stream` methods
 
 {PANEL:IsLoaded}
 
-To check if entity is attached to session, e.g. has been loaded previously, use `IsLoaded` method from `Advanced` session operations.
+To check if an entity is attached to a session, e.g. it has been loaded previously, use the `IsLoaded` method from the `Advanced` session operations.  
+  
+If you try to load a document that does not exist with the `Load` method, `IsLoaded` will return `true` because that document load has already been attempted.
```
reebhub, line 189: To check if an entity is `[tracked by a session](../../client-api/session/what-is-a-session-and-how-does-it-work#tracking-changes)`, e.g. because it has been loaded previously, use the `IsLoaded` method from the `Advanced` session operations.  

Note that `IsLoaded` will return `true` for a document even if you have previously attempted to load the document using the `Load` method and failed, because the document is now tracked by the session.

serezhub: I clarified it.

reebhub (approved):

PR #1456, opened 2022-03-03
RDoc-2092_ExplorationQueries - Exploration Queries

reebhub:


Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+  while the dataset is still held by the server.  
+
+* The **entire retrieved dataset** is scanned and filtered **without requiring or creating an 
+  index**, providing a way to conduct one-time explorations without creating an index that would 
+  have to be maintained by the cluster.  
+
+* You can filter the datasets retrieved by both **Index queries** and **Collection queries**.  
+
+* Exploration queries need to be used 
+  [catiously](../../indexes/querying/exploration-queries#when-should-exploration-queries-be-used),
```
arekpalinski, line 19: catiously -> cautiously ?

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+
+* You can filter the datasets retrieved by both **Index queries** and **Collection queries**.  
+
+* Exploration queries need to be used 
+  [catiously](../../indexes/querying/exploration-queries#when-should-exploration-queries-be-used), 
+  since the scanning and filtering of all the retrieved data without index makes them 
+  a **taxing operation** when large datasets are handled.  
+  We recommend that you -  
+    * **Limit** the number of records that an exploration query filters.  
+    * Use [where](../../indexes/querying/filtering) in recurring queries.
```
arekpalinski, line 24: Note that when using `where` then a query is always handled via index (if there is matching one then it will create a new auto index). When a query is handled by an index then `filter` will work as post query filtering.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+
+Exploration queries can be applied using -  
+
+* `Query.Filter`  
+* `DocumentQuery.Filter`  
+* RQL's `filter` keyword  
+
+The added filtering is parsed and executed by RavenDB's Javascript engine, 
+and can be implemented wherever Javascript is used, including within 
+Map/Reduce indexes and by subscription workers.
```
arekpalinski, line 48: `filter` is for querying. It's not applicable to JS indexing.

What was the indention of the second part of this sentence - "and can be implemented where .. "?

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+
+{PANEL: When Should Exploration Queries Be Used}
+
+`filter` can be applied to a Collection query, like in:  
+{CODE-BLOCK: javascript}
+from Employees as e 
+filter e.Address.Country = 'USA'
+{CODE-BLOCK/}
+
+it can also be applied to an Index query, like in:
```
arekpalinski, line 66: I'm not sure if "an Index query" is too big shortcut. Maybe let's put it explicitly "to a query handled by an index" and give two examples - one handing dynamic query via auto index (exactly what you have below) and one handling a query with the usage of an index `from index Users/BySomething filter Age > 30`

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+
+it can also be applied to an Index query, like in:  
+{CODE-BLOCK: javascript}
+from Employees as e 
+where e.Title = 'Sales Representative'  
+filter e.Address.Country = 'USA'
+{CODE-BLOCK/}
+
+In both cases, the entire retrieved dataset is scanned and filtered, 
+without using or creating an index.
```
arekpalinski, line 74: That's not correct - the second example will require and index since we have `where` clause used

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+
+In both cases, the entire retrieved dataset is scanned and filtered, 
+without using or creating an index.  
+This helps understand when exploration queries Should and should Not be used:  
+
+{INFO: }
+**Use** `filter` for an ad-hoc exploration of the retrieved dataset, that matches 
+no existing index and is not expected to be repeated much.  
+The dataset will be filtered without creating an unrequired index that the cluster 
+would continue updating from now on.
```
arekpalinski, line 81: It'd also mention that you do post query filtering on the server side when using an index

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+{INFO: }
+**Use** `filter` for an ad-hoc exploration of the retrieved dataset, that matches 
+no existing index and is not expected to be repeated much.  
+The dataset will be filtered without creating an unrequired index that the cluster 
+would continue updating from now on.  
+{INFO/}
+{WARNING: }
+Be aware, though, that when a large dataset is retrieved, like the whole collection in 
+the case of a collection query, exploring it all using `filter` would be taxing for the 
+server and cost the user a substantial waiting time.
```
arekpalinski, line 86: Regarding "taxing for the  server" - maybe make it clear it's about machine resource usage, especially the memory but also the CPU since we need to check the condition for each query result before we return it to the user

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+
+* **Limit** the number of records that an exploration query filters, e.g. -  
+  {CODE-BLOCK: javascript}
+  from Employees as e 
+where e.Title = 'Sales Representative'  
+filter e.Address.Country = 'USA'
+filter_limit 500 // limit the number of filtered records
+  {CODE-BLOCK/}
+* Use [where](../../indexes/querying/filtering) rather than `filter` for recurring filtering.  
+  `where` will use an index, creating it if necessary, to accelerate the filtering
```
arekpalinski, line 96: Here it is clear and correct.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+* `Query.Filter`  
+  {CODE-BLOCK:csharp}  
+  IRavenQueryable<T> Filter<T>(this IRavenQueryable<T> source, 
+                             Expression<Func<T, bool>> predicate, 
+                             int limit = int.MaxValue);
+  {CODE-BLOCK/}  
+
+    | Parameters | Type | Description |
+    | ---------- | ---- | ----------- |
+    | **source** | `IRavenQueryable<T>` | Query source |
```
arekpalinski, line 113: Maybe write that `Filter` is the extension method hence we have `this IRavenQueryable<T> source`

arekpalinski: BTW we multiple methods in our API is implemented like that e.g. `Search()` 

https://github.com/maciejaszyk/ravendb/blob/804a0a2e64b236a0ef12fce1059d27eb71a07b66/src/Raven.Client/Documents/LinqExtensions.cs#L929

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+Use `filter` with a collection query to scan and filter the entire collection.  
+{CODE-TABS}
+{CODE-TAB:csharp:Query exploration-queries_1.1@Indexes\Querying\ExplorationQueries.cs /}
+{CODE-TAB:csharp:DocumentQuery exploration-queries_1.2@Indexes\Querying\ExplorationQueries.cs /}
+{CODE-TAB:csharp:RawQuery exploration-queries_1.3@Indexes\Querying\ExplorationQueries.cs /}
+{CODE-TABS/}
+
+{WARNING: }
+Filtering a sizable collection will burden the server and prolong user waiting time.  
+Set a `limit` to restrict the number of filtered records.
```
arekpalinski, line 154: `filter_limit` not `limit`

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+{CODE-TAB:csharp:Query exploration-queries_2.1@Indexes\Querying\ExplorationQueries.cs /}
+{CODE-TAB:csharp:DocumentQuery exploration-queries_2.2@Indexes\Querying\ExplorationQueries.cs /}
+{CODE-TAB:csharp:RawQuery exploration-queries_2.3@Indexes\Querying\ExplorationQueries.cs /}
+{CODE-TABS/}
+
+---
+
+### With Projections (`select`)
+
+The filtered results can be projected using `select`.
```
arekpalinski, line 172: yes, it's the same as for any other query

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,219 @@
...
+
+// declare a Javascript function
+declare function titlePrefix(r, prefix) 
+{ 
+    // Add whatever filtering capabilities you like
+    return r.Title.startsWith(prefix)
+} 
+
+from Employees as e 
+where startsWith(e.HiredAt, '1993')
```
arekpalinski, line 200: This will cause index creation. I think we should drop it this example because users might be confused when they notice that it was handled by `Auto/Employees/ByHiredAt` index

Documentation/5.3/Samples/csharp/Raven.Documentation.Samples/Indexes/Querying/ExplorationQueries.cs
```diff
@@ -0,0 +1,108 @@
...
+                #region exploration-queries_1.3
+                result = session.Advanced.RawQuery<Employee>
+                    ("from Employees as e " +
+                       "filter e.Address.Country = 'USA' " +
+                       "filter_limit 500").SingleOrDefault();
+                #endregion
+                
+            }
+
+            // filter in an index query
```
arekpalinski, line 40: I think a example with a static index is missing. Just to show that it's possible

Documentation/5.3/Samples/csharp/Raven.Documentation.Samples/Indexes/Querying/ExplorationQueries.cs
```diff
@@ -0,0 +1,108 @@
...
+                #endregion
+            }
+
+            // filter and projection
+            using (var session = store.OpenSession())
+            {
+                #region exploration-queries_3.1
+                var emp1 = session
+                    .Query<Employee>()
+                    .Where(w => w.Title == "Sales Representative")
```
arekpalinski, line 76: Given that you have `Where` then it's not entirely "raw" exploration query - it will require an index creation

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
maciejaszyk: `by subscription workers.`
I do not know if it works.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
maciejaszyk: Agreed.

Some of tests were written this way to test the equality of `where` and `filter` sets.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
maciejaszyk: https://issues.hibernatingrhinos.com/issue/RavenDB-18133

Here is the ticket for testing that. For now, I think we shouldn't mention subscriptions in `filter` feature.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: fixed

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: made it a warning & added a link to the warning box it's explained below:
    
Use [where](../../indexes/querying/filtering) in recurring queries, so the filter would [use an index](../../indexes/querying/exploration-queries#limit-the-query-and-prefer--for-recurring-queries).

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: done
--> The added filtering is parsed and executed by RavenDB's Javascript engine.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: done -->

it can also be applied to queries handled by an index, e.g. -  

{CODE-BLOCK: javascript}
// dynamic query via an auto index  
from Employees as e 
where e.Title = 'Sales Representative'  
filter e.Address.Country = 'USA'
{CODE-BLOCK/}

{CODE-BLOCK: javascript}
// a query that uses an index explicitly  
from index 'Orders/ByCompany' 
filter Count > 10
{CODE-BLOCK/}

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: -->
Both in a collection query and in a query handled by an index, the entire retrieved dataset is scanned and filtered.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: added:
* You gain the ability to filter post-query results on the server side, for both collection queries and when an index was used.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: done

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: thx, added a link here to the previous mention in the top note

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: --> `Filter`, defined as an `IRavenQueryable` extension method

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: done

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: probably still worth mentioning it though
--> The filtered results can be projected using `select`, just like the results of any other query.

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub:
```
// declare a Javascript function
declare function titlePrefix(r, prefix) 
{ 
    // Add whatever filtering capabilities you like
    return r.Title.startsWith(prefix)
} 

from Employees as e 

// Filter using the function you've declared
filter titlePrefix(e, $prefix)
filter_limit 100
```

Documentation/5.3/Samples/csharp/Raven.Documentation.Samples/Indexes/Querying/ExplorationQueries.cs
reebhub: added
```
from index 'Orders/ByCompany' 
filter Count > 10
```

Documentation/5.3/Samples/csharp/Raven.Documentation.Samples/Indexes/Querying/ExplorationQueries.cs
reebhub: removed the `where` from the projection samples

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,237 @@
...
+  since scanning and filtering all the data retrieved by a query cost substantial 
+  [server resources and user waiting time](../../indexes/querying/exploration-queries#limit-the-query-and-prefer--for-recurring-queries) 
+  when large datasets are handled.  
+    {WARNING: }
+
+    We recommend that you -  
+
+    * **Limit** the number of records that an exploration query filters.  
+    * Use [where](../../indexes/querying/filtering) in recurring queries, 
+      so the filter would [use an index](../../indexes/querying/exploration-queries#limit-the-query-and-prefer--for-recurring-queries).
```
arekpalinski, line 29:
```suggestion
      so the query would [use an index](../../indexes/querying/exploration-queries#limit-the-query-and-prefer--for-recurring-queries).  

```

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,237 @@
...
+
+{CODE-BLOCK: javascript}
+// in a query that uses an index explicitly  
+from index 'Orders/ByCompany' 
+filter Count > 10
+{CODE-BLOCK/}
+
+Both in a collection query and in a query handled by an index, the entire retrieved 
+dataset is scanned and filtered.  
+This helps understand when exploration queries Should be used, why a Limit
```
arekpalinski, line 88:
```suggestion
This helps understand when exploration queries should be used, why a Limit 

```

arekpalinski, line 88: Should -> should?

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
```diff
@@ -0,0 +1,237 @@
...
+    | **limit** | `int ` | Limits the number of filtered records (Recommended) <br> Default: all retrieved records |
+
+* **RQL**  
+   * In an RQL query, use:  
+     The `filter` keyword, followed by the filtering condition.  
+     The `filter_limit` option, followed by the max number of records to filter.  
+   * E.g. -  
+     {CODE-BLOCK: javascript}
+     from Employees as e 
+where e.Title = 'Sales Representative'
```
arekpalinski, line 156: I'd drop `where e.Title = 'Sales Representative'` here to avoid index creation

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: done

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: done

Documentation/5.3/Raven.Documentation.Pages/indexes/querying/exploration-queries.markdown
reebhub: done

arekpalinski (approved):

PR #1469, opened 2022-03-31
RDoc-2115_RDoc-2116_Lazy

reebhub:
updated the lazy operations article with new lazy operations, updated the texts, edited and added samples

### DIFF
--- a/Documentation/5.1/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/HowTo/Lazy.cs
+++ b/Documentation/5.1/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/HowTo/Lazy.cs
@@ -63,16 +63,23 @@ namespace Raven.Documentation.Samples.ClientApi.Session.HowTo
 
                 }
 
+                string changeVector;
                 using (var session = store.OpenSession())
                 {
                     User user = new User();
-
+                    session.Store(user, "users/1-A");
+                    session.SaveChanges();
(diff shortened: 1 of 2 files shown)


Documentation/5.1/Samples/csharp/Raven.Documentation.Samples/ClientApi/Session/HowTo/Lazy.cs
```diff
@@ -0,0 +1,121 @@
...
+                    #region lazy_ConditionalLoad
+                    var lazy = session
+                        .Advanced
+                        .Lazily
+                        // Load only if the change vector has changed
+                        .ConditionalLoad<User>(user.Id, user.ChangeVector);
+                    
+                    var load = lazy.Value; // The operation will be executed here
+                    #endregion
+                }
```
garayx, line 79:
```suggestion
                string changeVector;
                using (var session = store.OpenSession())
                {
                    User user = new User();
                     session.Store(user, "users/1");
                     session.SaveChanges();
                     changeVector = session.Advanced.GetChangeVectorFor(user);
                }
                // New session which does not track our User entity
               using (var session = store.OpenSession())
                    #region lazy_ConditionalLoad
                    var lazy = session
                        .Advanced
                        .Lazily
                        // Load only if the change vector has changed
                        .ConditionalLoad<User>("users/1", changeVector);
                    
                    var load = lazy.Value; // The operation will be executed here
                    #endregion
                }

```

garayx (approved):

PR #1491, opened 2022-05-01
RDoc-2161-Created How to use DateOnly and TimeOnly article

serezhub:


Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,84 @@
...
+{PANEL: Convert and Use Date/TimeOnly Without Affecting Your Existing Data}
+
+[Static indexes](../../indexes/map-indexes) can process calculations on existing data and enable you to rapidly query it without changing your data.  
+
+{INFO: Ticks}
+To be able to use ticks with DateOnly or TimeOnly, you must create a **static index** that computes the conversion.  
+
+An auto-index will not convert DateOnly or TimeOnly into ticks, but will be able to convert data to strings.  
+You can also compare strings, though comparing ticks is usually much faster.  
+The index orders the strings according to the first digits, then the second, and so on.
```
maciejaszyk, line 53: This is the same algorithm used in radix sorting :)

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,84 @@
+# Client API: How to Use TimeOnly and DateOnly Types
+---
+
+{NOTE: }
+
+* RavenDB has introduced the ability to convert `DateTime` or strings written in date/time formats to .NET's
```
maciejaszyk, line 6: Not exactly. Microsoft in .NET6 introduced them and now RavenDB allows users to store and query on them in the same manner as DateTime.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,84 @@
...
+{INFO: Ticks}
+To be able to use ticks with DateOnly or TimeOnly, you must create a **static index** that computes the conversion.  
+
+An auto-index will not convert DateOnly or TimeOnly into ticks, but will be able to convert data to strings.  
+You can also compare strings, though comparing ticks is usually much faster.  
+The index orders the strings according to the first digits, then the second, and so on.  
+
+{INFO/}
+
+The index must be declared as DateOnly or TimeOnly.
```
maciejaszyk, line 57: Not the index, but a field inside the Index.
In this manner, we are sure that a user is giving a `DateOnly` or `TimeOnly` object (otherwise the compiler will throw an error). If not, there is also a possibility to call `AsDateOnly` or `AsTimeOnly` to give us information about `DateOnly` \ `TimeOnly`

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,84 @@
...
+You can also compare strings, though comparing ticks is usually much faster.  
+The index orders the strings according to the first digits, then the second, and so on.  
+
+{INFO/}
+
+The index must be declared as DateOnly or TimeOnly. 
+
+{CODE DateAndTimeOnlyIndexSample@ClientApi/HowTo/UseTimeOnlyAndDateOnly.cs /}
+
+Once the data has been converted inside the index, you can query on the converted data to find DateOnly or TimeOnly information.
```
maciejaszyk, line 61: Even if they are not converted you still can query on them. Main reason to create static index are the Ticks

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,84 @@
...
+{PANEL/}
+
+{PANEL: Convert and Use Date/TimeOnly Without Affecting Your Existing Data}
+
+[Static indexes](../../indexes/map-indexes) can process calculations on existing data and enable you to rapidly query it without changing your data.  
+
+{INFO: Ticks}
+To be able to use ticks with DateOnly or TimeOnly, you must create a **static index** that computes the conversion.  
+
+An auto-index will not convert DateOnly or TimeOnly into ticks, but will be able to convert data to strings.
```
maciejaszyk, line 51: It's not converting. Just treating data from documents as a normal string.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: Noted

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: done

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: Changed it to
`The index must have a field that declares the type as DateOnly or TimeOnly. `

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: Done.
This is the whole improved paragraph on that...
```
An auto-index will not convert DateOnly or TimeOnly into ticks, but will be data as strings.  
You can also compare strings, though comparing ticks is usually much faster and a static index needs to be used to do so.  
The index orders the strings according to the first digits, then the second, and so on (Radix Sorting).  
```

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: 1. Doesn't it save processing time/energy if the types are already converted?
2. How do they query on DateOnly if their data is DateTime? 
      a. In [this StackOverflow chain](https://stackoverflow.com/questions/22506930/how-to-query-datetime-field- 
          using-only-date-in-microsoft-sql-server) on that question there are like 20 answers and many of them look 
          too complex
3.  We don't have much documentation on querying ticks.  I found [this chain that discusses using ticks] 
    (https://groups.google.com/g/ravendb/c/rwMzQ96-zL4) in the google group, but there is no generic formula.  If 
    using ticks is much more efficient, it should be a part of every documentation article on querying (especially if it 
    has to do with querying date/time types)
        a. How would you advise a customer to query with ticks if their data is in DateTime?  I imagine that the first 
            step is to convert them in the index, but I may be wrong.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
maciejaszyk: Where?

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+
+* RavenDB has introduced the ability to convert `DateTime` or strings written in date/time formats to .NET's 
+  `DateOnly` or `TimeOnly` types without slowing down queries and while leaving your existing data as is.  
+   * Use `AsDateOnly` or `AsTimeOnly` in a static index (SEE EXAMPLES BELOW)
+   * `AsDateOnly` and `AsTimeOnly` automatically convert strings to ticks for faster querying.  
+
+* We convert the types in [static indexes](../../indexes/map-indexes) so that the conversions and calculations are done behind the scenes
+  and the data is ready for fast queries. [See sample index below.](../../client-api/how-to/using-timeonly-and-dateonly#convert-and-use-date/timeonly-without-affecting-your-existing-data)).  
+
+* It is also possible to covert via patch or query, but it is [far more efficient](../../client-api/how-to/using-timeonly-and-dateonly#convert-and-use-date/timeonly-without-affecting-your-existing-data)
```
maciejaszyk, line 18: It's not possible. You can modify data through the patch to get a compatible format but honestly, it's not a good idea. I think this part should be deleted.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+  According to [Microsoft .NET Blog](https://devblogs.microsoft.com/dotnet/date-time-and-time-zone-enhancements-in-net-6/#the-timeonly-type)
+  TimeOnly is ideal for scenarios such as recurring meeting times, daily alarm clock times, 
+  or the times that a business opens and closes each day of the week.
+  * See [their usage examples here.](https://devblogs.microsoft.com/dotnet/date-time-and-time-zone-enhancements-in-net-6/#the-timeonly-type)
+
+{PANEL/}
+
+{PANEL: Convert and Use Date/TimeOnly Without Affecting Your Existing Data}
+
+It is possible to covert via patch or query, but it is far more efficient to do so in a [static index](../../indexes/map-indexes).
```
maciejaszyk, line 47: Same, you can transform through the patch, but not query. Projections are not converting anything for Querying/Indexing.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+
+{PANEL/}
+
+{PANEL: Convert and Use Date/TimeOnly Without Affecting Your Existing Data}
+
+It is possible to covert via patch or query, but it is far more efficient to do so in a [static index](../../indexes/map-indexes). 
+
+   * Patch changes your existing data, which may cause problems with existing logic that uses that data.
+   * Queries become slow if they are defined with demanding processes.  
+     Also, queries will need to process all of your data every time.
```
maciejaszyk, line 51: ? what do you mean

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+
+   * Patch changes your existing data, which may cause problems with existing logic that uses that data.
+   * Queries become slow if they are defined with demanding processes.  
+     Also, queries will need to process all of your data every time. 
+   * [Static indexes](../../indexes/indexing-basics) process new data in the background, 
+    including calculations and converstions to DateOnly/TimeOnly values, which can be used as ticks,  
+    so that the data is ready at query time when you [query the index](../../indexes/querying/basics#example-iv---querying-a-specified-index).  
+       * These indexes do all of the calculations on the entire dataset defined the first time they run, and then they only need to 
+         process changes in data, which is far less expensive than running the calculations and conversions on the entire dataset 
+         each time with a query.
```
maciejaszyk, line 57: At this point, I think an explanation about how Indexes works are not necessary.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+
+{INFO: Ticks}
+To be able to use ticks as DateOnly or TimeOnly, you must create a **static index** that computes the conversion from strings.  
+
+An auto-index will not convert DateOnly or TimeOnly into ticks, but will process strings as strings.  
+By defining the query that creates an auto-index that orders the strings, you can also compare strings, 
+though comparing ticks is usually much faster.  
+Strings can be ordered according to the first digits, then the second, and so on (Radix Sorting).
```
maciejaszyk, line 65: not necessary.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+
+Now you can query on the converted data to find DateOnly or TimeOnly information.  
+
+For example, the following query will find all of the entries that occured between 15:00 and 17:00 
+without considering the date.
+
+{CODE TimeOnly_between15-17@ClientApi/HowTo/UseTimeOnlyAndDateOnly.cs /}
+
+**Querying on Ticks**  
+Because the query is on a static index, you can rapidly query on ticks if you choose.
```
maciejaszyk, line 181: We are prioritizing ticks over strings in queries. So if the user created ticks (or it has happened automatically), the user cannot decide how we do the query internally.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+    var element = session.Query<DateOnlyItem, StringAsDateOnlyConversion>().Where(item => item.DateOnlyField == today).As<StringItem>().Single();
+}
+{CODE-BLOCK/}
+
+---
+
+#### Converting `DateTime` with minimal cost
+
+The following generic sample is a map index that converts `DateTime` into `DateOnly` in the index instead of doing expensive 
+conversions repetetively in queries.
```
maciejaszyk, line 123: Don't get it.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+Strings can be ordered according to the first digits, then the second, and so on (Radix Sorting).  
+
+RavenDB automatically makes data converted via `AsDateOnly` or `AsTimeOnly` available as ticks.  
+{INFO/}
+
+### Use `AsDateOnly` or `AsTimeOnly` in a static index to convert strings or DateTime types  
+
+#### Converting Strings with minimal cost
+
+The following generic sample is a map index where `AsDateOnly` converts the string `item.StringDateOnlyField` into `DateTime`.
```
maciejaszyk, line 74: `DateOnly` ;)

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,197 @@
...
+RavenDB automatically makes data converted via `AsDateOnly` or `AsTimeOnly` available as ticks.  
+{INFO/}
+
+### Use `AsDateOnly` or `AsTimeOnly` in a static index to convert strings or DateTime types  
+
+#### Converting Strings with minimal cost
+
+The following generic sample is a map index where `AsDateOnly` converts the string `item.StringDateOnlyField` into `DateTime`.  
+
+The conversion automatically makes the DateTime type available as ticks.
```
maciejaszyk, line 76: DateTime data always is saved with ticks. Doing `AsDateOnly` we are trimming the Time part.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: I changed it to `You can now convert DateTime or strings written in date/time formats to .NET's DateOnly or TimeOnly types...`

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: Removed

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: Removed.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: That is what I thought we discussed on Zoom.  I removed this section because even if it was right, we don't need to tell users what not to do...only what to do.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: It seems important because it explains the reasoning for using the index to create a converted copy of the data.  Most articles that I've read about converting `DateTime` to `DateOnly` discuss using .NET's `FromDateTime` method.  We offer a different method that is used in the index.  

It's a very brief summary, with a link to learn more.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: Good catch. thanks!

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: I previously thought that indexing is a better alternative to using queries, which [seems to be a common way](https://stackoverflow.com/questions/22506930/how-to-query-datetime-field-using-only-date-in-microsoft-sql-server).

Clearly the way we're presenting is much better for many reasons, but if using a query is common practice, isn't it a good idea to explain why we use indexes instead?

serezhub (commented):
Does it make sense to use this?  It isn't in the #region, and we needed to get `waitforindexing(store)` to work.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -41,25 +42,30 @@ These two new C# types are available from .NET 6.0+ (RavenDB 5.3+).
...
+RavenDB automatically converts strings into ticks via `AsDateOnly` or `AsTimeOnly`.  
 {INFO/}
 
-### Use `AsDateOnly` or `AsTimeOnly` in a static index to convert strings or DateTime types  
+### Use `AsDateOnly` or `AsTimeOnly` in a static index to convert strings or DateTime
+
+* [Converting Strings to DateOnly or TimeOnly](../../client-api/how-to/using-timeonly-and-dateonly#converting-strings-with-minimal-cost)
+* [Converting DateTime to DateOnly or TimeOnly](../../client-api/how-to/using-timeonly-and-dateonly#converting--with-minimal-cost)
 
 #### Converting Strings with minimal cost
```
maciejaszyk, line 70: Maybe we will concatenate strings & DateTime conversion examples into one example? They are very similar.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
maciejaszyk:
```How do they query on DateOnly if their data is DateTime?
a. In [this StackOverflow chain](https://stackoverflow.com/questions/22506930/how-to-query-datetime-field-
using-only-date-in-microsoft-sql-server) on that question there are like 20 answers and many of them look
too complex```

You cannot compare them directly from C# code, for example ;) BUT you can use RQL. Your DateTime will be converted into Ticks (as usual).

File: Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown (hunk 9 above)
Comment (maciejaszyk) [reply Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown]: Internally, we are doing `FromDateTime` ;) 
https://github.com/ravendb/ravendb/blob/016785f4d3d96485af3e972f64ba3eaff05a7e7e/src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs#L462

Comment (maciejaszyk) [reply Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown]: We are calling our own converters when the user has stored data as strings already.

File: Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown (hunk 10 above)
Comment (maciejaszyk) [reply Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown]: My mistake, sorry.

File: Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown (hunk 12 above)
Comment (maciejaszyk) [reply Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown:91]: This is T-SQL (SQL) and RavenDB is NoSQL. RavenDB requires an index (exploration query is the exception).

File: Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown (hunk 15 above)
Comment (serezhub) [reply Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown:70]: Does it make sense to combine them into one code sample?  
They use different variables and they reference different indexes.  
Also, the user won't have both strings and DateTime, so if the sample is combined, will that throw an error?

File: Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown [hunk 16]
```diff
@@ -0,0 +1,136 @@
...
+{PANEL: Convert and Use Date/TimeOnly Without Affecting Your Existing Data}
+
+In RavenDB, we offer conversion of types in static indexes with the methods [AsDateOnly or AsTimeOnly](../../client-api/how-to/using-timeonly-and-dateonly#use--or--in-a-static-index-to-convert-strings-or-datetime-types).
+
+* [Static indexes](../../indexes/indexing-basics) process new data in the background, 
+  including calculations and converstions to DateOnly/TimeOnly values, which can be used as ticks,  
+  so that the data is ready at query time when you [query the index](../../indexes/querying/basics#example-iv---querying-a-specified-index).  
+    * These indexes do all of the calculations on the entire dataset that you define the first time they run, and then they only need to 
+      process changes in data. This is far less expensive than running the calculations and conversions on the entire dataset 
+      each time you query and it doesn't affect your existing data.
```
maciejaszyk, line 52: As I said before, we are always creating an index for search. I think you can delete this part. @arekpalinski what do you think?

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
maciejaszyk: ` but will be data as strings. ` -> `but will index data as strings` maybe?

+
`Strings can be ordered according to the first digits, then the second, and so on (Radix Sorting).  `

I'm not sure if we want to add that :)

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: I see

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: I took out the part about queries

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: I took out the last sentence about queries.

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
maciejaszyk: Yes

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: Changed to `Now you can use ticks to query on the converted data in the index.  `

But, now if the point is about using ticks, does the following query use ticks?
```
    public void QuerySampleForIndexWithDateTimeAndDateOnly()
        {
                using (var store = new DocumentStore())
                {
                    using var session = store.OpenSession();
                    #region TimeOnly_between15-17
                    var after = new TimeOnly(15, 00);
                    var before = new TimeOnly(17, 00);
                    var result = session
                    .Query<DateAndTimeOnly>()
                    .Where(i => i.TimeOnly > after && i.TimeOnly < before)
                    .ToList();
                    #endregion
                }
        }
```

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: I took out the sentence about radix

maciejaszyk (approved):

maciejaszyk:
LGTM

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,131 @@
...
+  According to [Microsoft .NET Blog](https://devblogs.microsoft.com/dotnet/date-time-and-time-zone-enhancements-in-net-6/#the-timeonly-type)
+  TimeOnly is ideal for scenarios such as recurring meeting times, daily alarm clock times, 
+  or the times that a business opens and closes each day of the week.
+  * See [their usage examples here.](https://devblogs.microsoft.com/dotnet/date-time-and-time-zone-enhancements-in-net-6/#the-timeonly-type)
+
+{PANEL/}
+
+{PANEL: Convert and Use Date/TimeOnly Without Affecting Your Existing Data}
+
+In RavenDB, we offer conversion of types in static indexes with the methods [AsDateOnly or AsTimeOnly](../../client-api/how-to/using-timeonly-and-dateonly#use--or--in-a-static-index-to-convert-strings-or-datetime).
```
reebhub, line 45: RavenDB offers..

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,131 @@
...
+
+* [Static indexes](../../indexes/indexing-basics) process new data in the background, 
+  including calculations and converstions to DateOnly/TimeOnly values, which can be used as ticks,
```
reebhub, line 48: * fix "converstions" (& run it through Grammarly)
* the line is cut in the wrong place

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -0,0 +1,131 @@
...
+
+In RavenDB, we offer conversion of types in static indexes with the methods [AsDateOnly or AsTimeOnly](../../client-api/how-to/using-timeonly-and-dateonly#use--or--in-a-static-index-to-convert-strings-or-datetime).
+
+* [Static indexes](../../indexes/indexing-basics) process new data in the background, 
+  including calculations and converstions to DateOnly/TimeOnly values, which can be used as ticks,  
+  so that the data is ready at query time when you [query the index](../../indexes/querying/basics#example-iv---querying-a-specified-index).  
+    * These indexes do all of the calculations on the entire dataset that you define the first time they run, and then they only need to 
+      process changes in data. 
+
+{INFO: Ticks}
```
reebhub, line 53: you don't explain what ticks are anywhere, including in the box called "ticks"

reebhub (changes requested):

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: done

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: done

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
serezhub: done

reebhub (approved):

Documentation/5.3/Raven.Documentation.Pages/client-api/how-to/using-timeonly-and-dateonly.markdown
```diff
@@ -94,18 +100,23 @@ Once the converted data is available in the static index, you can inexpensively
 
 {CODE IndexConvertsDateTimeWithAsDateOnlySample@ClientApi/HowTo/UseTimeOnlyAndDateOnly.cs /}
 
+{NOTE: }
+RavenDB doesn't look for DateOnly or TimeOnly as default types during indexing
+so the variables must be wrapped in AsDateDonly() or AsTimeOnly() explicitly.
```
maciejaszyk, line 105: Why did you paste it 3 times?

serezhub: There are three different sections with code samples.  It would be out of context in the intro section because it's about the code syntax, so I needed to put the note in each section.

Documentation/5.3/Samples/csharp/Raven.Documentation.Samples/ClientApi/HowTo/UseTimeOnlyAndDateOnly.cs
```diff
@@ -151,6 +163,8 @@ public class IndexEntry
             public DateAndTimeOnlyIndex()
             {
                 Map = dates => from date in dates
+                               // RavenDB doesn't look for DateOnly or TimeOnly as default types during indexing
+                               // so they need to be called explicitly.
```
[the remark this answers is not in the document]
serezhub: done

PR #1519, opened 2022-06-20
RDoc-2135-Put-Command-5.0+ changed line 39 in sample file

serezhub:
I only changed one line - line 39 in the sample file

### DIFF
--- a/Documentation/5.0/Raven.Documentation.Pages/client-api/commands/documents/put.dotnet.markdown
+++ b/Documentation/5.0/Raven.Documentation.Pages/client-api/commands/documents/put.dotnet.markdown
@@ -1,16 +1,26 @@
 # Commands: Documents: Put
 
+---
+
+{NOTE: }
 **Put** is used to insert or update a document in a database.
+
+In this page: 
+
+* [Syntax](../../../client-api/commands/documents/put#syntax)
+* [Example](../../../client-api/commands/documents/put#example)
+
+{NOTE/}
 
 ## Syntax
 
 {CODE put_interface@ClientApi\Commands\Documents\Put.cs /}
 
-| Parameters | | |
+| Parameters | Type | Description |
 | ------------- | ------------- | ----- |
 | **id** | string | unique ID under which document will be stored |
 | **changeVector** | string | Entity changeVector, used for concurrency checks (`null` to skip check) |
-| **document** | BlittableJsonReaderObject | The document to store. You may use `session.Advanced.JsonConverter.ToBlittable(doc, docInfo);` to convert your entity to a `BlittableJsonReaderObject` |
+| **document** | BlittableJsonReaderObject | The document to store. You may use `session.Advanced.JsonConverter.ToBlittable(doc, docInfo);` to convert your entity to a `BlittableJsonReaderObject`. |
 
 ## Example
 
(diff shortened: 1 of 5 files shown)


Documentation/5.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Commands/Documents/Put.cs
```diff
@@ -0,0 +1,47 @@
...
+                var doc = new Category
+                {
+                    Name = "My category",
+                    Description = "My category description"
+                };
+                var docInfo = new DocumentInfo
+                {
+                    Collection = "Categories"
+                };
+                var blittableDoc = session.Advanced.JsonConverter.ToBlittable(doc, docInfo);
```
reebhub, line 39: * you updated the sample without updating the method's prototype and explanations about it
* please check in which version the new method replaced the old one, go through the documentation and make sure we document only the correct method.

Documentation/5.0/Raven.Documentation.Pages/client-api/commands/documents/put.dotnet.markdown
```diff
@@ -0,0 +1,25 @@
...
+
+## Syntax
+
+{CODE put_interface@ClientApi\Commands\Documents\Put.cs /}
+
+| Parameters | | |
+| ------------- | ------------- | ----- |
+| **id** | string | unique ID under which document will be stored |
+| **changeVector** | string | Entity changeVector, used for concurrency checks (`null` to skip check) |
+| **document** | BlittableJsonReaderObject | The document to store. You may use `session.Advanced.EntityToBlittable.ConvertEntityToBlittable` to convert your entity to a `BlittableJsonReaderObject` |
```
reebhub, line 13: you updated the method usage in the sample but left the old explanations here. (and in other copies of the article, e.g. in version 5.2)

reebhub (changes requested):

Documentation/5.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Commands/Documents/Put.cs
serezhub: Added //comments to explain the sample.

We're supposed to change it in 5.0 plus.  I copied this into 5.0 and checked that it bubbles up to 6.0.

Documentation/5.0/Raven.Documentation.Pages/client-api/commands/documents/put.dotnet.markdown
serezhub: I see.  Good catch. Done

Documentation/5.0/Samples/csharp/Raven.Documentation.Samples/ClientApi/Commands/Documents/Put.cs
reebhub: not sure what you changed, the 5.2 explanation remained the same.
please go through the versions as many times as you need and finish it properly.

serezhub: Done.

reebhub (approved):

PR #1592, opened 2023-01-10
RDoc-2041_ShardingOverview2023_2 added Sharding root folder and overview

reebhub:


Documentation/6.0/Raven.Documentation.Pages/sharding/overview.markdown
```diff
@@ -0,0 +1,250 @@
...
+{WARNING/}
+
+{PANEL/}
+
+{PANEL: Resharding}
+
+**Resharding** is a process in which the content of a bucket 
+(e.g. bucket 100000 in shard **1**) is reallocated to another 
+bucket (e.g. bucket 350000 in shard **2**) to maintain a balanced 
+database in which all shards handle about the same volume of data.
```
aviv86, line 213: that's not accurate
the content of the bucket moves to another shard, but the bucket number doesn't change
it's the **Bucket Allocations** that changes - the mapping from shards to the bucket ranges they're assigned to
so bucket 100000 that previously belonged to shard 1 will now belong to shard 2 (after we finish moving all the data)

reebhub: done

aviv86 (approved):

PR #1622, opened 2023-03-17
RDoc-2375_shardingAndQueries - indexing & querying pages

reebhub:


Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -3,24 +3,182 @@
...
+  
 {NOTE/}
 
 ---
 
-{PANEL: }
+{PANEL: Filtering Results on a Sharded Database}
+
+Query results can be filtered either on each shard, or on the orchestrator 
+machine after the results are collected by it from all shards.
```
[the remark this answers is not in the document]
arekpalinski: The problem is that in sharded env users should use `filter` in order to do conditional queries against map-reduce indexes if a field isn't a group by key.

It's this issue: https://issues.hibernatingrhinos.com/issue/RavenDB-18164/Query-against-a-map-reduce-index-should-get-all-the-results#focus=Comments-67-397839.0-0

And the usage of `filter` and applying it on the orchestrator is a solution for that.

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -3,24 +3,182 @@
...
+"Loading a document inside a projection from a map-reduce index isn't supported".  
+
+Unlike map-reduce index projections, projections of queries that use no index 
+and projections of map indexes **can** load a document, providing that the document 
+is [on this shard](../sharding/querying#unsupported-querying-features).  
+
+| Projection | Can load Document | Conditions |
+| ---------- | ----------------- | ---------- |
+| Query Projection | Yes | The document is on this shard |
+| Map Index Projection | Yes | The document is on this shard |
```
arekpalinski, line 74: When a loaded document isn't on the shard we return just `null`. We don't load it under the covers from another shard where it exists

https://issues.hibernatingrhinos.com/issue/RavenDB-17889/Load-inside-a-projection

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -3,24 +3,182 @@
...
+  and sets a limit to restrict the retrieved dataset to 3 results.  
+  This **will** restrict the retrieved dataset to the set limit.  
+  {CODE-BLOCK:csharp}
+                    var queryResult = session.Query<UserMapReduce.Result, UserMapReduce>()
+                        .OrderBy(x => x.Name)
+                        .Take(3) // this limit will apply while retrieving the items
+                        .ToList();
+  {CODE-BLOCK/}
+  
+* The third query sorts the results **not** by a reduce key but by by
```
arekpalinski, line 121: doubled "by"

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -3,24 +3,182 @@
...
+                        .OrderBy(x => x.Sum)
+                        .Take(3) // this limit will only apply after retrieving all items
+                        .ToList();
+  {CODE-BLOCK/}
+    
+  {NOTE: }
+  Note that retrieving all the results from all shards, either 
+  by setting no limit or by setting a limit based on a computation 
+  as demonstrated above, may cause the retrieval of a large amount 
+  of data and extend memory, CPU, and bandwidth usage.
```
arekpalinski, line 137: I wonder if we shouldn't recommend that the best performance will be offered by queries which does filtering on the reduce keys `.Where(x => x.Name == "Foo")`? CC @grisha-kotler

arekpalinski (commented):
1. I think we're missing a section explaining how Includes are handled. We generate additional to relevant shards from the orchestrator - so this might be important factor affecting query execution time

2. Also there isn't any info about paging. Especially deep paging might be expensive.

3. We need to document this ability: 
https://issues.hibernatingrhinos.com/issue/RavenDB-18159/Add-an-option-to-select-the-shards-for-query-in-the-client

Documentation/6.0/Raven.Documentation.Pages/sharding/indexing.markdown
```diff
@@ -3,24 +3,88 @@
...
+  * [Unsupported Indexing Features](../sharding/indexing#unsupported-indexing-features)  
 
 {NOTE/}
 
 ---
 
-{PANEL: }
+{PANEL: Indexing}
+
+Indexing and indexes under each shard are basically similar to indexing
```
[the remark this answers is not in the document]
reebhub: done

Documentation/6.0/Raven.Documentation.Pages/sharding/indexing.markdown
```diff
@@ -3,24 +3,88 @@
...
+## Unsupported Indexing Features
+
+Unsupported or yet-unimplemented indexing features include: 
+
+* **Rolling index deploymeny**  
+  [Rolling index deploymeny](../indexes/rolling-index-deployment) 
+  is not supported on a sharded database.  
+* **Loading documents from other shards**  
+  Loading a document during indexing is possible only if the document 
+  resides on the indexed shard.
```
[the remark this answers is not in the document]
reebhub: done

Documentation/6.0/Raven.Documentation.Pages/sharding/indexing.markdown
```diff
@@ -3,24 +3,88 @@
...
+and indexes on a non-sharded database.  
+As each shard holds and manages a unique dataset, indexing is performed 
+per-shard and indexes are stored only on the shard that created and uses them.  
+
+## Map Reduce Indexes on a Sharded Database
+
+A map reduce index will reduce the query results **twice**.  
+First, the results will be reduced by each shard.  
+Then, the results will be collected by the orchestrator from 
+all shards and reduced again.
```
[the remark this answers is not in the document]
reebhub: When a map reduce index is used, the query results are reduced **twice**:  
The first reduction is made per-shard, when the query is executed on each 
shard and the shard reduces the retrieved results by the index.  
The second reduction is made by the orchestrator over all query results 
after it collects them from all shards.

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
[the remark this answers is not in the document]
reebhub: yeah i moved this section out of the map reduce panel by mistake, put it back in, thx

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
reebhub: there's a reference to the "not supported" section where it's explained, i enhanced the clarification there & changed it from ignoring the request to returning `null`.

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
reebhub: thx, fixed

Documentation/6.0/Raven.Documentation.Pages/sharding/indexing.markdown
```diff
@@ -3,24 +3,88 @@
...
+
+A map reduce index will reduce the query results **twice**.  
+First, the results will be reduced by each shard.  
+Then, the results will be collected by the orchestrator from 
+all shards and reduced again.  
+
+{NOTE: }
+Note that this behavior is different from the behavior of 
+map reduce indexes in non-sharded databases, and may yield 
+different results.
```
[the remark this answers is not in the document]
reebhub: removed

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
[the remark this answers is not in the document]
reebhub: "The problem is that".. added a recommendations section

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
reebhub: added many changes to the map reduce sections and a `where` vs `filter` recommendations section

Documentation/6.0/Raven.Documentation.Pages/sharding/indexing.markdown
```diff
@@ -3,24 +3,90 @@
...
+Map reduce indexes on a sharded database are used to reduce data both over each 
+shard during indexation, and on the orchestrator machine each time a query uses them.  
+
+1. **Reduction by each shard during indexation**  
+   Similarly to non-sharded databases, when shards index their data they reduce 
+   the results by map reduce indexes.  
+2. **Reduction by the orchestrator during queries**  
+   When a query is executed over map reduce indexes the orchestrator 
+   distributed the query to the shards, collects and combines the results, 
+   and then reduces the entire dataset.
```
arekpalinski, line 38: I think that "entire dataset"  might be a bit misleading. It's about the results returned from the shards.

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -1,26 +1,295 @@
...
+  * [Include](../sharding/querying#include)  
+  * [Unsupported Querying Features](../sharding/querying#unsupported-querying-features)  
+  * [Querying a Specific Shard](../sharding/querying#querying-a-specific-shard)  
+  
 {NOTE/}
 
 ---
+{PANEL: Querying in a Sharded Database}
+
+From a user's' point of view, querying a sharded RavenDB database
```
arekpalinski, line 27: Too many `'`?

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -1,26 +1,295 @@
...
+  When a query that includes a `where` clause is sent to a **sharded database**, 
+  filtering is applied **per-shard**, over each shard's database.  
+  
+     This presents us with the following problem:  
+     The filtering that runs on each shard takes into account only the sales that were 
+     recorded on that shard.  
+     If a certain product was sold 4000 times on each shard, the query demonstrated 
+     above will filter this product out on each shard, even though its total sales 
+     far exceed 5000.  
+     To solve this problem, the role of the `filter` command is altered on sharded databases.
```
arekpalinski, line 99: I wonder if we shouldn't mention here that filtering with the usage of `where` on group by fields is okay and it doesn't have this issue

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -1,26 +1,295 @@
...
+  the database and then scanned by the filtering condition.  
+  On the pros side, this mechanism allows us to perform queries 
+  like the one we had a problem with when we used `where`.  
+  The below query will indeed return all the products that 
+  were sold at least 5000 times, no matter how their sales 
+  are divided between the shards.  
+     {CODE-BLOCK:JSON}
+     from index 'Products/Sales'
+     filter TotalSales >= 5000
+     {CODE-BLOCK/}
```
arekpalinski, line 132: Maybe mention that in order to reduce the number of results returned by each shard users can additionally add (if that makes sense in their case) `where` condition with a group by field

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
```diff
@@ -1,26 +1,295 @@
...
+  a conditional query field like [Total or Sum](../sharding/querying#orderby-in-a-map-reduce-index) field.  
+* When using `filter`, set a [limit](../indexes/querying/exploration-queries#usage) if possible.  
+* When `filter` is needed, use `where` first to minimize the dataset that 
+  needs to be transferred from the shards to the orchestrator and scanned 
+  by `filter` over the orchestrator machine. E.g. -  
+  {CODE-BLOCK:JSON}
+  var emp = session.Query<Employee>()
+  .Where(w => w.Title == "Sales Representative")
+  .Filter(f => f.Address.Country == "USA", limit: 500)
+  .SingleOrDefault();
```
arekpalinski, line 234: This is not a query against map-reduce index

Documentation/6.0/Raven.Documentation.Pages/sharding/indexing.markdown
reebhub: -> 
When a query is executed over map reduce indexes the orchestrator 
distributes the query to the shards, collects and combines the results, 
and then reduces them again.

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
reebhub: To solve this problem, the role of the `filter` command is [altered on sharded databases](../sharding/querying#section-1).  
     {NOTE: }
     Using `where` raises no problem, and is actually [recommended](../sharding/querying#vs--recommendations), 
     when the filtering is done [over a GroupBy field](../sharding/querying#orderby-in-a-map-reduce-index).  
     {NOTE/}

Documentation/6.0/Raven.Documentation.Pages/sharding/querying.markdown
reebhub:
```
from index 'Products/Sales'
where Category = 'categories/7-A'
filter TotalSales >= 5000
```

PR #1750, opened 2024-02-15
RDoc-2294 Getting Started: Add info about the Cloud option

hiddenshadow21:


Documentation/4.0/Raven.Documentation.Pages/start/getting-started.markdown
```diff
@@ -55,6 +55,13 @@ If you are interested in hosting the server in a Docker container, please read o
 
 {NOTE/}
 
+{NOTE: Running in RavenDB Cloud}
+
+If you want to test RavenDB without manual setup try [RavenDB Cloud](https://cloud.ravendb.net).
+We offer one free instance per customer. For more information, please read our [dedicated article](../cloud/cloud-overview).
+
+{NOTE/}
+
```
reebhub, line 64: The cloud interface is documented only from version 4.2 and up, so the second link won't work (there's no cloud folder or cloud overview markdown) and i'm not sure what part of the note is relevant. 
The course of action should probably be - 
- Remove the note from this file and its 4.1 version
- copy the 4.1 version of getting-started.markdown to version 4.2 and add the note there

reebhub (approved):

PR #1789, opened 2024-03-07
RDoc-2658 - [Python] Client API > Session > Querying > Document Query > What is a Document Query [Replace C# samples]

reebhub:
[RDoc-2658](https://issues.hibernatingrhinos.com/issue/RDoc-2658/Python-Client-API-Session-Querying-Document-Query-What-is-a-Document-Query-Replace-C-samples)


poissoncorp (approved):
makes perfect sense, thanks :]

PR #1922, opened 2024-10-09
RDoc-2441 Fix URL for removing node from cluster topology

Danielle9897:
**Related issue:**
https://issues.hibernatingrhinos.com/issue/RDoc-2441/Incorrect-URL-for-node-removing


ArieSLV:
I don't have any comments, but I'd like to clarify: will these changes also be applied to the corresponding pages for versions 5.4 through 7.0?

ArieSLV (approved):

Danielle9897:
@ArieSLV 
Yes, the changes applied in this PR to the existing "Cluster Topology" markdown articles will be visible across all higher versions. 
This is due to the fall-through/bubble mechanism. 
For example, if the **v7.0** directory doesn’t have a Cluster Topology article, 
the system will look for the article in lower versions, and the highest available version (**v5.3** in this case) will be displayed.

ArieSLV:
Good to know.
Thank you for your explanation 🙏

PR #2043, opened 2025-05-22
RDoc-3343 Vector search: Querying multiple vector fields

Danielle9897:
**Related issue:**
https://issues.hibernatingrhinos.com/issue/RDoc-3343/Vector-search-Querying-multiple-vector-fields

### DIFF
--- a/Documentation/7.0/Raven.Documentation.Pages/ai-integration/vector-search/vector-search-using-dynamic-query.dotnet.markdown
+++ b/Documentation/7.0/Raven.Documentation.Pages/ai-integration/vector-search/vector-search-using-dynamic-query.dotnet.markdown
@@ -21,6 +21,7 @@
   * [Dynamic vector search - exact search](../../ai-integration/vector-search/vector-search-using-dynamic-query#dynamic-vector-search---exact-search)
   * [Quantization options](../../ai-integration/vector-search/vector-search-using-dynamic-query#quantization-options)
   * [Querying vector fields and regular data in the same query](../../ai-integration/vector-search/vector-search-using-dynamic-query#querying-vector-fields-and-regular-data-in-the-same-query)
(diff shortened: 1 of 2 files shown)


Documentation/7.0/Raven.Documentation.Pages/ai-integration/vector-search/vector-search-using-dynamic-query.dotnet.markdown
```diff
@@ -655,6 +656,45 @@ and (vector.search(embedding.text(Name), $searchTerm, 0.75, 25))
...
+  * Or companies with a _Name_ similar to "dairy"
+
+  {CODE-TABS}
+  {CODE-TAB:csharp:DocumentQuery vs_27@AiIntegration\VectorSearch\VectorSearchUsingDynamicQuery.cs /}
+  {CODE-TAB:csharp:DocumentQuery_async vs_27_async@AiIntegration\VectorSearch\VectorSearchUsingDynamicQuery.cs /}
+  {CODE-TAB:csharp:RawQuery vs_28@AiIntegration\VectorSearch\VectorSearchUsingDynamicQuery.cs /}
+  {CODE-TAB:csharp:RawQuery_async vs_28_async@AiIntegration\VectorSearch\VectorSearchUsingDynamicQuery.cs /}
+  {CODE-TAB-BLOCK:sql:RQL}
+from "Companies"
+where
```
ayende, line 678: An important aspect here is that `vector.search()` will only return a set amount of results (by default - IIRC, is 16.
We evaluate each clause independently, so you may have this:

```
vector.search(embedding.text(Name), $searchTerm1, 0.78)
  and
  vector.search(embedding.text(Address.Country), $searchTerm2, 0.82)
```

Where the first clause gives `[1,2,3...16]` results and the second gives [18,... 34]` results. And there won't be a match between them.

However, if we will continue deeper in the first clause, we'll get to `18` which would be a match.

The problem is that we must run the HNSW on each vector independently, and the size matters a lot.

ayende (approved):

PR #2068, opened 2025-08-01
RDoc-3361 Added deployment script

Lwiel:


scripts/deploy.ps1
```diff
@@ -0,0 +1,189 @@
...
+  -CloudFrontDistributionId <string>  # distribution to invalidate (if omitted, no invalidation)
+  -Versions <string[]>                # explicit list of RavenDB versions (e.g. 6.0,6.2,7.0,7.1)
+Version convenience switches (alias-based, can be combined):
+  -6.0  -6.2  -7.0  -7.1
+------------------------------------------------------------------------------!#>
+
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory=$true)]
+    [string]$AwsRegion,
```
gregolsky: region is going to come from AWS_DEFAULT_REGION env var

scripts/deploy.ps1
```diff
@@ -0,0 +1,189 @@
...
+    Session   = $Env:AWS_SESSION_TOKEN
+}
+ThrowIfEmpty $EnvCreds.AccessKey "AWS_ACCESS_KEY_ID environment variable not set"
+ThrowIfEmpty $EnvCreds.Secret    "AWS_SECRET_ACCESS_KEY environment variable not set"
+
+# export region for downstream CLI calls
+$Env:AWS_REGION = $AwsRegion
+
+Write-Host "Using AWS region $AwsRegion and bucket $S3BucketName" -ForegroundColor Cyan
+
```
gregolsky: we just need to check if they are there, no need to load it into any var like EnvCreds - aws CLI will manage

scripts/deploy.ps1
```diff
@@ -0,0 +1,187 @@
...
+    # AWS CLI check
+    if (-not (Get-Command aws -ErrorAction SilentlyContinue)) {
+        throw "AWS CLI v2 is required but was not found in PATH."
+    }
+    Write-Host "AWS CLI version: $(aws --version)" -ForegroundColor Gray
+
+    # Ensure Docusaurus CLI is available
+    if (-not (Get-Command docusaurus -ErrorAction SilentlyContinue)) {
+        Write-Host "Docusaurus CLI not found – installing locally (no-save)..." -ForegroundColor Yellow
+        npm install --no-save @docusaurus/core@latest @docusaurus/cli@latest | Out-Null
```
gregolsky: can we use npm ci?

static/robots.txt
```diff
@@ -0,0 +1,17 @@
+# robots.txt for RavenDB Documentation
+User-agent: *
+Allow: /
+
+# Block duplicate or low-value paths
+Disallow: /test/
+Disallow: /wip/
+
+# AI crawlers you want to exclude
```
poissoncorp: actually we want to allow AI training on these docs. Change to allow for all agents not only GPTBot

static/robots.txt
```diff
@@ -0,0 +1,17 @@
+# robots.txt for RavenDB Documentation
+User-agent: *
+Allow: /
+
+# Block duplicate or low-value paths
```
poissoncorp: we don't have such paths - no need for that

static/robots.txt
```diff
@@ -0,0 +1,17 @@
+# robots.txt for RavenDB Documentation
+User-agent: *
```
poissoncorp, line 2: I think we could just allow to any agent - AI, robots, etc.. Ensure it's not risky. 

This way we could have 

allow-all
sitemap

and it's done

scripts/build_whats_new.py
```diff
@@ -0,0 +1,194 @@
...
+
+import requests
+from markdownify import markdownify as md
+
+# ============================================================================
+# Configuration & paths
+# ============================================================================
+
+API_KEY: str | None = os.environ.get("RAVENDB_API_KEY")
+API_BASE_URL = "https://api.web.ravendb.net/api/v1/documentation/whats-new"
```
poissoncorp, line 56: @gregolsky when testing, you should add test. to the URL here

scripts/build_whats_new.py
```diff
@@ -0,0 +1,194 @@
...
+
+def mdx_heading(entry: Dict[str, Any]) -> str:
+    """Create a level‑2 MDX heading from an API entry."""
+    date_str = datetime.strptime(entry["buildDate"], API_DATE_FMT).strftime("%Y/%m/%d")
+    return f"## {entry['version']} - {date_str}\n\n"
+
+
+def mdx_block(entry: Dict[str, Any]) -> str:
+    """Full MDX chunk for a single changelog entry (heading + body)."""
+    return mdx_heading(entry) + convert_html_to_markdown(entry["changelogHtml"])
```
gregolsky, line 142: We can talk with the API team to add us source markdown field, which won't feel then like getting strawberries out of a milkshake. cc @lukaszdobrzynski 

However we can go live with this, not a blocker.

scripts/deploy.ps1
```diff
@@ -0,0 +1,150 @@
...
+Write-Host "Syncing to s3://$S3BucketName/ …" -ForegroundColor Cyan
+aws s3 sync $BuildDir "s3://$S3BucketName/" --delete
+if ($LASTEXITCODE) { throw 'aws s3 sync failed' }
+
+# ---------------------------------------------------------------------------
+# 6. Invalidate CloudFront (optional)
+# ---------------------------------------------------------------------------
+if ($CloudFrontDistributionId) {
+    Write-Host "Invalidating CloudFront distribution $CloudFrontDistributionId" -ForegroundColor Cyan
+    aws cloudfront create-invalidation --distribution-id $CloudFrontDistributionId --paths '/*' | Out-Null
```
gregolsky, line 146: later on if we know what version changed we could us that knowledge to invalidate proper path

static/robots.txt
gregolsky: robots.txt is only for "nice" robots looking at signs on the road, note

scripts/build_whats_new.py
```diff
@@ -0,0 +1,194 @@
...
+# ============================================================================
+# REST helpers
+# ============================================================================
+
+def get_api_page(branch: str, page: int = 1) -> Dict[str, Any]:
+    """Return a single paginated payload from the Documentation API."""
+    response = requests.get(
+        API_BASE_URL,
+        headers={
+            "Authorization": f"Bearer {API_KEY}",
```
gregolsky, line 90: Well, you won't be able to do it like this I'm afraid. This token I supplied you with is short-lived and to get a new one, you'd need some more details and ceremony. To get it going let's download these items, make what's new files and push it to the repo for the time being.

This one is a no-go in the current shape.

gregolsky (changes requested):
what's new script won't do

scripts/build_whats_new.py
gregolsky: Actually, we can make it work. @poissoncorp Please arrange a short one with @lukaszdobrzynski - we'd need a new secret.

poissoncorp (approved):

docusaurus.config.ts
```diff
@@ -53,10 +42,9 @@ const config: Config = {
               path: "/7.1"
             }
           }
-          // Please change this to your repo.
           // Remove this to remove the "edit this page" links.
           // editUrl:
-          //  'https://github.com/facebook/docusaurus/tree/main/packages/create-docusaurus/templates/shared/',
+          //  'https://github.com/ravendb/docs/new-docs/main/'
```
poissoncorp, line 47: remember to change it while deploying @Lwiel

PR #2140, opened 2025-09-28
RDoc-2293 [Per-feature reorganization] Compare-Exchange articles

Danielle9897:
**Related issues:**
* https://issues.hibernatingrhinos.com/issue/RDoc-2293/Per-feature-reorganization-Compare-Exchange-articles
* https://issues.hibernatingrhinos.com/issue/RDoc-3309/Compare-Exchange-includes-should-not-override-already-tracked-compare-exchange-values-in-the-session
* https://issues.hibernatingrhinos.com/issue/RDoc-3141/Improve-documentation-around-working-with-compare-exchange-under-session

---

@haludi =>   **C# files to review + start & configuration**:
```
docs/compare-exchange/start.mdx
docs/compare-exchange/content/_overview-csharp.mdx
docs/compare-exchange/content/_create-cmpxchg-items-csharp.mdx
docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx
docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
docs/compare-exchange/content/_update-cmpxchg-item-csharp.mdx
docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx
docs/compare-exchange/content/_include-compare-exchange-items-csharp.mdx
docs/compare-exchange/content/_indexing-compare-exchange-values-csharp.mdx
docs/compare-exchange/content/_cmpxchg-in-dynamic-queries-csharp.mdx
docs/compare-exchange/content/_cmpxchg-item-expiration-csharp.mdx
docs/compare-exchange/content/_atomic-guards-csharp.mdx
docs/compare-exchange/configuration.mdx 
```
---

@M4xymm => **Node.js files to review**:
```
docs/compare-exchange/content/_create-cmpxchg-items-nodejs.mdx
docs/compare-exchange/content/_get-cmpxchg-item-nodejs.mdx
docs/compare-exchange/content/_get-cmpxchg-items-nodejs.mdx
docs/compare-exchange/content/_update-cmpxchg-item-nodejs.mdx
docs/compare-exchange/content/_delete-cmpxchg-items-nodejs.mdx
docs/compare-exchange/content/_include-compare-exchange-items-nodejs.mdx
docs/compare-exchange/content/_indexing-compare-exchange-values-nodejs.mdx
docs/compare-exchange/content/_cmpxchg-in-dynamic-queries-nodejs.mdx
docs/compare-exchange/content/_cmpxchg-item-expiration-nodejs.mdx
docs/compare-exchange/content/_atomic-guards-nodejs.mdx
```
---

@kalczur => **tsx file to review**
This allows us to add a basic card without an image.
It's a good solution until an image is provided/generated, or in cases where adding images is too noisy or crowded in the UI.
```
src/components/Common/Card.tsx
```
---

Content for the other languages (Java, PHP, Python) was simply copied over - no code review is currently needed.
Articles for these languages should be handled separately in dedicated PRs.
This is summarized in the table **“Status of article coverage per client language”** in the issue description:
https://issues.hibernatingrhinos.com/issue/RDoc-2293/Per-feature-reorganization-Compare-Exchange-articles

### DIFF
--- a/docs/compare-exchange/content/_overview-csharp.mdx
+++ b/docs/compare-exchange/content/_overview-csharp.mdx
@@ -239,13 +239,16 @@ public void PrintWork()
     try
     {
         // Do some work for the duration that was set (TimeSpan.FromMinutes(20)).
-        // Don't exceed the duration, otherwise the resource won't be available for someone else.
-        
-        // In a distributed system (unlike a multi-threaded app), a process may fail or exit unexpectedly 
-        // before releasing the resource - never reaching the 'finally' block.
-        
-        // To prevent the resource from remaining locked indefinitely, 
-        // we use a timeout to expire the reservation.
+        //
+        // In a distributed system (unlike a multi-threaded app), a process may crash or exit unexpectedly  
+        // without releasing the resource it reserved (i.e. never reaching the 'finally' block).  
+        // This can leave the resource locked indefinitely.
+        //
+        // To prevent that, each reservation includes a timeout (TimeSpan.FromMinutes(20)).  
+        // If the process fails or exits, the resource becomes available again once the timeout expires.
+        //
+        // Important: Ensure the work completes within the timeout period.  
+        // If it runs longer, another client may acquire the same resource at the same time.
     }
     finally
     {
--- a/docs/compare-exchange/content/_overview-java.mdx
+++ b/docs/compare-exchange/content/_overview-java.mdx
@@ -223,14 +223,17 @@ public void printWork() throws InterruptedException \{
     long reservationIndex = lockResource(store, "Printer/First-Floor", Duration.ofMinutes(20));
 
     try \{
-        // Do some work for the duration that was set (Duration.ofMinutes(20)).
-        // Don't exceed the duration, otherwise the resource won't be available for someone else.
-        
-        // In a distributed system (unlike a multi-threaded app), a process may fail or exit unexpectedly 
-        // before releasing the resource - never reaching the 'finally' block.
-        
-        // To prevent the resource from remaining locked indefinitely, 
-        // we use a timeout to expire the reservation.
+        // Do some work for the duration that was set (TimeSpan.FromMinutes(20)).
+        //
+        // In a distributed system (unlike a multi-threaded app), a process may crash or exit unexpectedly  
+        // without releasing the resource it reserved (i.e. never reaching the 'finally' block).  
+        // This can leave the resource locked indefinitely.
+        //
+        // To prevent that, each reservation includes a timeout (TimeSpan.FromMinutes(20)).  
+        // If the process fails or exits, the resource becomes available again once the timeout expires.
+        //
+        // Important: Ensure the work completes within the timeout period.  
+        // If it runs longer, another client may acquire the same resource at the same time.
     \} finally \{
         releaseResource(store, "Printer/First-Floor", reservationIndex);
     \}
--- a/docs/compare-exchange/content/_overview-php.mdx
+++ b/docs/compare-exchange/content/_overview-php.mdx
@@ -236,14 +236,17 @@ class CompareExchangeSharedResource
         $reservationIndex = $this->lockResource($this->store, "Printer/First-Floor", Duration::ofMinutes(20));
 
         try \{
-            // Do some work for the duration that was set (Duration::ofMinutes(20))).
-            // Don't exceed the duration, otherwise the resource won't be available for someone else.
-            
-            // In a distributed system (unlike a multi-threaded app), a process may fail or exit unexpectedly 
-            // before releasing the resource - never reaching the 'finally' block.
-            
-            // To prevent the resource from remaining locked indefinitely, 
-            // we use a timeout to expire the reservation.
+            // Do some work for the duration that was set (TimeSpan.FromMinutes(20)).
+            //
+            // In a distributed system (unlike a multi-threaded app), a process may crash or exit unexpectedly  
+            // without releasing the resource it reserved (i.e. never reaching the 'finally' block).  
+            // This can leave the resource locked indefinitely.
+            //
+            // To prevent that, each reservation includes a timeout (TimeSpan.FromMinutes(20)).  
+            // If the process fails or exits, the resource becomes available again once the timeout expires.
+            //
+            // Important: Ensure the work completes within the timeout period.  
+            // If it runs longer, another client may acquire the same resource at the same time.
         \} finally \{
             $this->releaseResource($this->store, "Printer/First-Floor", $reservationIndex);
         \}
--- a/docs/compare-exchange/content/_overview-python.mdx
+++ b/docs/compare-exchange/content/_overview-python.mdx
@@ -203,14 +203,17 @@ def print_work() -> None:
 
     try:
         ...
-        # Do some work for the duration that was set (timedelta(minutes=20)).
-        # Don't exceed the duration, otherwise the resource won't be available for someone else.
-        
-        # In a distributed system (unlike a multi-threaded app), a process may fail or exit unexpectedly 
-        # before releasing the resource - never reaching the 'finally' block.
-        
-        # To prevent the resource from remaining locked indefinitely, 
-        # we use a timeout to expire the reservation.
+        # Do some work for the duration that was set (TimeSpan.FromMinutes(20)).
+        #
+        # In a distributed system (unlike a multi-threaded app), a process may crash or exit unexpectedly  
+        # without releasing the resource it reserved (i.e. never reaching the 'finally' block).  
+        # This can leave the resource locked indefinitely.
+        #
+        # To prevent that, each reservation includes a timeout (TimeSpan.FromMinutes(20)).  
+        # If the process fails or exits, the resource becomes available again once the timeout expires.
+        #
+        # Important: Ensure the work completes within the timeout period.  
+        # If it runs longer, another client may acquire the same resource at the same time.
     finally:
         release_resource(store, "Printer/First-Floor", reservation_index)
 


M4xymm (approved):

docs/compare-exchange/start.mdx
```diff
@@ -0,0 +1,294 @@
...
+* Compare-exchange is a RavenDB feature for storing atomic, cluster-wide **key-value pairs** where each key is a globally unique identifier in the database.
+  Items are versioned and managed at the cluster level.
+  
+* Compare-exchange provides a built-in consensus mechanism ideal for safe coordination and global consistency  
+  in distributed environments, allowing you to:
+  * Enforce global uniqueness (e.g., prevent duplicate usernames or emails).
+  * Assign work to a single client or reserve a resource once.
+  * Handle concurrency safely, without external services or custom locking logic.
+  
+* Compare-exchange items can also serve as a place to store shared data that lives outside of the documents -
```
haludi: I don't understand this section.

docs/compare-exchange/start.mdx
```diff
@@ -0,0 +1,294 @@
...
+  in distributed environments, allowing you to:
+  * Enforce global uniqueness (e.g., prevent duplicate usernames or emails).
+  * Assign work to a single client or reserve a resource once.
+  * Handle concurrency safely, without external services or custom locking logic.
+  
+* Compare-exchange items can also serve as a place to store shared data that lives outside of the documents -  
+  useful for values that are reused, referenced globally, or managed independently of any single document.
+  
+* Key Characteristics of a compare-exchange item:
+  * Cluster-wide - Visible and consistent across all nodes.
```
haludi: We worth to specify "all nodes in the database group"

docs/compare-exchange/start.mdx
```diff
@@ -0,0 +1,294 @@
...
+* Simple to implement and works seamlessly across the cluster.
+
+</Admonition>
+
+<Admonition type="note" title="">
+
+### Add safety to cluster-wide transactions
+
+* RavenDB uses **compare-exchange items** behind the scenes to protect cluster-wide transactions.  
+  When using cluster-wide sessions, RavenDB automatically creates internal compare-exchange entries called Atomic Guards
```
haludi: When using cluster-wide sessions to handle documents, RavenDB automatically uses compare-exchange to enforce the modification to be atomic.

docs/compare-exchange/start.mdx
```diff
@@ -0,0 +1,294 @@
...
+  to coordinate concurrent access and prevent conflicting writes.
+
+* ✅ Why compare-exchange?  
+  It provides a Raft-based coordination mechanism that ensures consistency and safety during multi-node transactions.
+
+* How it works:
+  * When you store or update a document in a cluster-wide session,  
+    RavenDB generates an atomic guard with a Raft-based version.
+  * This guard tracks the document’s state across the cluster.
+  * If another session modifies the document (or its state changes due to replication),
```
haludi: "or its state changes due to replication"
I don't think this is relevant - implementation details.
What does it add to the user understanding?

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+---
+    
+## What compare-exchange items are
+
+Compare-exchange items are key/value pairs where the key serves as a unique value across your database.
+
+* Each compare-exchange item contains: 
+  * **A key** - A unique string identifier in the database scope.
+  * **A value** - Can be any object (a number, string, array, or any valid JSON object).
```
haludi: Can be any **object**
I think it should be "value"

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+* Each compare-exchange item contains: 
+  * **A key** - A unique string identifier in the database scope.
+  * **A value** - Can be any object (a number, string, array, or any valid JSON object). 
+  * **Metadata** - Optional data that is associated with the compare-exchange item. Must be a valid JSON object.  
+    For example, the metadata can be used to set expiration time for the compare-exchange item.  
+    Learn more in [compare-exchange expiration](../compare-exchange/cmpxchg-expiration).  
+  * **Raft index** - The compare-exchange item's version.  
+    Any change to the value or metadata will increase this number.  
+
+* Creating and modifying a compare-exchange item is an atomic, thread-safe [compare-and-swap](https://en.wikipedia.org/wiki/Compare-and-swap)
```
haludi: It is similar to the compare-and-swap operation in a process, but in RavenDB, it serves a distributed system rather than a multi-threaded system.

docs/compare-exchange/start.mdx
Danielle9897: how about:
``` 
* Compare-exchange items can also store shared or global values that aren't tied to a specific document.  
  For example: system-wide configuration flags, shared settings, or reusable identifiers.
  ````

File: docs/compare-exchange/start.mdx (hunk 2 above)
Comment (Danielle9897) [reply docs/compare-exchange/start.mdx]: done

File: docs/compare-exchange/start.mdx (hunk 3 above)
Comment (Danielle9897) [reply docs/compare-exchange/start.mdx]: =>
```
### Add safety to cluster-wide transactions

* When using cluster-wide sessions to handle documents, RavenDB automatically creates internal compare-exchange items,
  called [atomic guards](../compare-exchange/atomic-guards), to enforce atomic document modifications.  
  These items coordinate access and prevent conflicting writes across nodes.  

* ✅ Why compare-exchange?  
  It provides a Raft-based coordination mechanism that ensures consistency and safety during multi-node transactions.

* How it works:
  * When you store or update a document in a cluster-wide session,  
    RavenDB creates an atomic guard to track the document’s version across the cluster.
  * If another session modifies the document in the meantime,  
    your transaction fails with a `ConcurrencyException`, ensuring data consistency.

* This protects you from:
  * Writing over documents that were modified by other sessions.
  * Acting on stale data in a distributed environment.
  * Violating ACID guarantees in multi-node clusters.

* You don’t need to manage these guards manually -   
  RavenDB handles everything automatically when you use a session in cluster-wide mode.
  ```

docs/compare-exchange/start.mdx
Danielle9897: removed

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+* **Document Store Operations**  
+  You can create and manage compare-exchange items using _document store_ operations.  
+  For example, see [Create items using a store operation](../compare-exchange/create-cmpxchg-items#create-items-using-a-store-operation).
+
+* **Cluster-Wide Sessions**  
+  You can create and manage compare-exchange items from within a [Cluster-Wide session](../client-api/session/cluster-transaction/overview#cluster-wide-transaction-vs-single-node-transaction).  
+  For example, see [Create items using a cluster-wide session](../compare-exchange/create-cmpxchg-items#create-items-using-a-cluster-wide-session).     
+  When using a cluster-wide session, the compare-exchange item is created as part of the cluster-wide transaction.  
+  If the session fails, the item creation also fails, and none of the nodes in the database group will store the new compare-exchange item.
```
haludi: Something is not clear here.
If the session "fails".
What can make it fail?
We only validate the cluster command in the cluster transaction.
Of course, the request can fail for a network issue, but that is not relevant here, and it is more complicated than true or false.

I would write something like, "like a transaction, all of the commands succeed or none of them."

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+  For details about what is and what isn't replicated in [What is Replicated](../server/ongoing-tasks/external-replication#what-is-replicated).
+    
+---
+    
+## Example I - Email address reservation  
+
+The following example shows how to use compare-exchange to create documents with unique values.  
+The scope is within the database group on a single cluster. 
+
+Compare-exchange items are not externally replicated to other databases.
```
haludi: It looks like it shouldn't be here.
Is it a duplicate of the above?

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+User user = new User
+{
+    Email = email
+};
+
+using (IDocumentSession session = store.OpenSession())
+{
+    session.Store(user);
+    // At this point, the user document has an Id assigned
```
haludi: The object (C# object) has an ID assigned.
The document in RavenDB wasn't created at this point.

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+
+    // Try to reserve a new user email 
+    // Note: This operation takes place outside of the session transaction, 
+    //       It is a cluster-wide reservation
+    CompareExchangeResult<string> cmpXchgResult
+        = store.Operations.Send(
+            new PutCompareExchangeValueOperation<string>(email, user.Id, 0));
```
haludi, line 154: Maybe worth mentioning that using 0 here is the key to say we want the operation to succeed only if this PUT creates the compare-exchange.

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+                new PutCompareExchangeValueOperation<SharedResource>(resourceName, resource, 0));
+
+        if (putResult.Successful)
+        {
+            // resourceName wasn't present - we managed to reserve
+            return putResult.Index;
+        }
+
+        // At this point, Put operation failed - someone else owns the lock or lock time expired
+        if (putResult.Value.ReservedUntil < now)
```
haludi, line 282: I think it is worth mentioning why we do that, and we don't just try to create it with 0.
There can be a case where the process that locked the resource crashed or anything else, so it will not release it.

docs/compare-exchange/content/_overview-csharp.mdx
```diff
@@ -0,0 +1,345 @@
...
+    // We have 2 options here:
+    // deleteResult.Successful is true - we managed to release resource
+    // deleteResult.Successful is false - someone else took the lock due to timeout 
+}
+```
+</TabItem>
+
+---
+
+## Example III - Ensuring unique values without using compare exchange
```
haludi: I am not familiar with this option, and I don't think it should work.  
I tested it, and I can make it fail.

Cluster of 3:  
Node C is taken down.  
A document 'user/1' is created with a cluster-wide transaction.  
Nodes A and B are taken down.  
Node C is started.  
A new application tries to load 'user/1' with a cluster-wide transaction.  
The result is 'null'.

docs/compare-exchange/start.mdx
haludi: What is the advantage over a normal document?

docs/compare-exchange/content/_overview-csharp.mdx
Danielle9897: done

docs/compare-exchange/content/_overview-csharp.mdx
Danielle9897: =>
```
Creating and modifying a compare-exchange item follows the same principle as the compare-and-swap operation
in multi-threaded systems, but in RavenDB, this concept is applied to a distributed environment across 
multiple nodes instead of within a single multi-threaded process.  
  ```

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 7 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_overview-csharp.mdx]: =>
```
When using a cluster-wide session, the compare-exchange item is created as part of the cluster-wide transaction.  
Like any transaction, all operations either succeed as a group or are rolled back.  
If the transaction is not committed, the new compare-exchange item will not be stored on any node in the database group.
```

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 8 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_overview-csharp.mdx]: removed

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 9 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_overview-csharp.mdx]: =>
```
// At this point, the user object has a document ID assigned by the session.
```

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 10 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_overview-csharp.mdx:154]: done

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 11 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_overview-csharp.mdx:282]: =>
```
// At this point, someone else owns the resource.
// But if their reservation has expired (e.g., the process crashed and never released it),
// we can try to take the lock by overwriting the value using the current index.
```

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 12 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_overview-csharp.mdx]: 1. Removed this example.
2. Added this new section instead:

```
## Why not use regular documents to enforce uniqueness

* You might consider storing a document with a predictable ID (for example, _phones/123456_) as a way to enforce uniqueness, 
  and then checking for its existence before allowing another document to use the same value.

* While this might work in a single-node setup or with external replication,
  it does not reliably enforce uniqueness in a clustered environment.
 
* If a node was not part of the cluster when the document was created, it might not be aware of its existence when it comes back online. 
  In such cases, attempting to load the document on this node may return _null_, leading to duplicate values being inserted.

* To reliably enforce uniqueness across all cluster nodes, you must use compare-exchange items,
  which are designed for this purpose and ensure global consistency.
```

File: docs/compare-exchange/start.mdx (hunk 1 above)
Comment (Danielle9897) [reply docs/compare-exchange/start.mdx]: I removed this bullet from _start.mdx_ and moved it to _overview.mdx_, where this is now explained.

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 11 above)
Comment (haludi) [reply docs/compare-exchange/content/_overview-csharp.mdx:282]: I think it is still unclear.
The motivation is not clear.

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 12 above)
Comment (haludi) [reply docs/compare-exchange/content/_overview-csharp.mdx]: Why mention that in the first place?  
It only makes things more confusing instead of explaining.

File: docs/compare-exchange/content/_create-cmpxchg-items-csharp.mdx [hunk 13]
```diff
@@ -0,0 +1,564 @@
...
+    }
+    ```
+    </TabItem>
+    </Tabs>
+
+---
+
+## Create item using a store operation
+  
+* Use the `PutCompareExchangeValueOperation` [store operation](../client-api/operations/what-are-operations) to create a compare-exchange item independently, without opening a session.
```
haludi: It would be good to emphasize here that the 0 index is what defines the operation as creating rather than modifying.  
Basically, this operation is used for both.

docs/compare-exchange/content/_create-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,564 @@
...
+```csharp
+public PutCompareExchangeValueOperation(
+    string key, T value, long index, IMetadataDictionary metadata = null)
+```
+</TabItem>
+
+| Parameter    | Type                  | Description |
+|--------------|-----------------------|-------------|
+| **key**      | `string`              | <ul><li>A unique identifier in the database scope.</li><li>Can be up to 512 bytes.</li></ul> |
+| **value**    | `T`                   | <ul><li>A value to be saved for the specified _key_.</li><li>Can be any object (number, string, array, or any valid JSON object).</li></ul> |
```
haludi: Can be any value

docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx
```diff
@@ -0,0 +1,514 @@
...
+                 key: "user1-name@example.com",
+                 value: "users/1" 
+         );
+         
+         // Optionally, add some metadata:
+         itemToCreate.Metadata["email-type"] = "work email";
+         
+         session.SaveChanges();
+    
+        // Get the compare-exchange item:
```
haludi, line 72: It’s better to do that in a different session.  
As you wrote, the session tracks the compare-exchange, so it will take it from its state.

docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx
```diff
@@ -0,0 +1,514 @@
...
+                key: "user1-name@example.com",
+                value: "users/1"
+            );
+            
+        // Optionally, add some metadata:
+        itemToCreate.Metadata["email-type"] = "work email";    
+            
+        await asyncSession.SaveChangesAsync();
+        
+        // Get the compare-exchange item:
```
haludi, line 118: It’s better to do that in a different session.  
As you wrote, the session tracks the compare-exchange, so it will take it from its state.

docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx
```diff
@@ -0,0 +1,514 @@
...
+        new SessionOptions { TransactionMode = TransactionMode.ClusterWide }))
+    {     
+         CompareExchangeValue<string> itemToCreate = 
+             session.Advanced.ClusterTransaction.CreateCompareExchangeValue(
+                 key: "user1-name@example.com",
+                 value: "users/1" 
+         );
+         session.SaveChanges();
+         
+        // Get the compare-exchange item lazily:
```
haludi, line 164: It’s better to do that in a different session.  
As you wrote, the session tracks the compare-exchange, so it will take it from its state.

docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx
```diff
@@ -0,0 +1,514 @@
...
+    {
+        CompareExchangeValue<string> itemToCreate =
+            asyncSession.Advanced.ClusterTransaction.CreateCompareExchangeValue(
+                key: "user1-name@example.com",
+                value: "users/1"
+            );
+            
+        await asyncSession.SaveChangesAsync();
+         
+        // Get the compare-exchange item lazily:
```
haludi, line 205: It’s better to do that in a different session.  
As you wrote, the session tracks the compare-exchange, so it will take it from its state.

docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,454 @@
...
+    {
+        // Define the list of keys of the compare-exchange items to retrieve
+        var keys = new[] { "employees/1", "employees/2", "customers/2", "non-existing-key" };
+            
+        // Call 'GetCompareExchangeValues', pass the list of keys 
+        Dictionary<string, CompareExchangeValue<string>> items =
+            session.Advanced.ClusterTransaction.GetCompareExchangeValues<string>(keys);
+        
+        // Check results    
+        Debug.Assert(items.Count == 4); // 4 keys were requested
```
haludi: The count of the result is always the same as the requested.  
This code is part of our testing, not something the user should do.

docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,454 @@
...
+    {
+        // Define the list of keys of the compare-exchange items to retrieve
+        var keys = new[] { "employees/1", "employees/2", "customers/2", "non-existing-key" };
+            
+        // Call 'GetCompareExchangeValues', pass the list of keys 
+        Dictionary<string, CompareExchangeValue<string>> items =
+            await asyncSession.Advanced.ClusterTransaction.GetCompareExchangeValuesAsync<string>(keys);
+        
+        // Check results    
+        Debug.Assert(items.Count == 4); // 4 keys were requested
```
haludi: The count of the result is always the same as the requested.
This code is part of our testing, not something the user should do.

docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,454 @@
...
+    var keys = new[] { "employees/1", "employees/2", "customers/2", "non-existing-key" };
+        
+    // Define the get compare-exchange items operation, pass the unique item key
+    var getCmpXchgItemsOp = new GetCompareExchangeValuesOperation<string>(keys);
+
+    // Execute the operation by passing it to Operations.Send
+    Dictionary<string, CompareExchangeValue<string>> items = store.Operations.Send(getCmpXchgItemsOp);
+    
+    // Check results    
+    Debug.Assert(items.Count == 4); // 4 keys were requested
```
haludi: The count of the result is always the same as the requested.
This code is part of our testing, not something the user should do.

docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,454 @@
...
+        
+    // Define the get compare-exchange items operation, pass the unique item key
+    var getCmpXchgItemsOp = new GetCompareExchangeValuesOperation<string>(keys);
+
+    // Execute the operation by passing it to Operations.SendAsync
+    Dictionary<string, CompareExchangeValue<string>> items = await 
+        store.Operations.SendAsync(getCmpXchgItemsOp);
+    
+    // Check results    
+    Debug.Assert(items.Count == 4); // 4 keys were requested
```
haludi: The count of the result is always the same as the requested.
This code is part of our testing, not something the user should do.

docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,454 @@
...
+        var keys = new[] { "employees/1", "employees/2", "customers/2", "non-existing-key" };
+            
+        // Call 'Lazily.GetCompareExchangeValues', pass the list of keys 
+        var lazyItems =
+            session.Advanced.ClusterTransaction.Lazily.GetCompareExchangeValues<string>(keys);
+        
+        Dictionary<string, CompareExchangeValue<string>> items = lazyItems.Value;
+        
+        // Check results    
+        Debug.Assert(items.Count == 4); // 4 keys were requested
```
haludi: The count of the result is always the same as the requested.
This code is part of our testing, not something the user should do.

docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,454 @@
...
+        var keys = new[] { "employees/1", "employees/2", "customers/2", "non-existing-key" };
+            
+        // Call 'Lazily.GetCompareExchangeValues', pass the list of keys 
+        var lazyItems = asyncSession.Advanced.ClusterTransaction
+            .Lazily.GetCompareExchangeValuesAsync<string>(keys);
+                    
+        Dictionary<string, CompareExchangeValue<string>> items = await lazyItems.Value;
+        
+        // Check results    
+        Debug.Assert(items.Count == 4); // 4 keys were requested
```
haludi: The count of the result is always the same as the requested.
This code is part of our testing, not something the user should do.

docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,454 @@
...
+        // Call 'GetCompareExchangeValues', pass:
+        // * startWith: The common key prefix
+        // * start:  The start position (optional, default is 0)
+        // * pageSize: Max items to get (optional, default is 25)
+        Dictionary<string, CompareExchangeValue<string>> items =
+            session.Advanced.ClusterTransaction.GetCompareExchangeValues<string>(
+                startsWith: "employees", start: 0, pageSize: 10);
+    
+        // Results will include only compare-exchange items with keys that start with "employees"
+        Debug.Assert(items.Count == 3); // There are 3 keys with the "employees" prefix
```
haludi: I don't think `Debug.Assert` is a good way to demonstrate the usage of the value.
But that is only my opinion

docs/compare-exchange/content/_update-cmpxchg-item-csharp.mdx
```diff
@@ -0,0 +1,236 @@
...
+
+## Update compare-exchange item using a store operation
+
+* Use `PutCompareExchangeValueOperation` to **update the _value_ and/or _metadata_** of an existing compare-exchange item.
+  This operation is also used to create new compare-exchange items, see [Create compare-exchange item](../compare-exchange/create-cmpxchg-items).
+
+* To perform an update, provide:
+  * The existing key
+  * A new value and/or metadata
+  * The latest index (version) of the item as last seen by the client.
```
haludi: I think it will be more understandable if you write something like "the expected version/index"

docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,365 @@
...
+  
+* **Delete items by expiration**:  
+  Compare-exchange items can also be deleted by adding an expiration date to them.  
+  Learn more in [Compare-exchange expiration](../compare-exchange/cmpxchg-expiration).
+  
+* **Compare-exchange tombstones**:  
+  Whenever a compare-exchange item is deleted, a compare-exchange tombstone item is created for it.  
+  Tombstones are used to indicate to other RavenDB processes that the item was deleted, so they can react accordingly.  
+  E.g., indexes referencing the deleted item will update themselves to remove those references.  
+  The tombstones themselves are not removed immediately - RavenDB uses an internal cleaner task to periodically remove tombstones that are eligible for deletion.
```
haludi: Very minor.
Saying "The tombstones themselves are not removed immediately" makes it sound too important.  
The customer doesn’t really need to care about that.  
All he needs to know is that it’s a background process.  
Anyway, it’s not something he interacts with directly.

docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,365 @@
...
+  * [Syntax](../compare-exchange/delete-cmpxchg-items#syntax)
+
+</Admonition>
+
+---
+
+## Delete compare-exchange item using a cluster-wide session
+  
+* Delete compare-exchange items using a cluster-wide session when you want the deletion to be part of a transaction committed via `SaveChanges()`. 
+  This is suitable if you want to include compare-exchange deletions and document changes in the same transaction.
```
haludi: Minor
It is also suitable if you want to delete and put a compare exchange as one transaction.
In general, this is to do multiple commands in one transaction.

docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,365 @@
...
+  Learn more about cluster-wide sessions in [Cluster transactions - overview](../client-api/session/cluster-transaction/overview).   
+
+* Use `DeleteCompareExchangeValue()` to register the deletion of an existing compare-exchange item in the session.  
+  The item will be deleted as part of the cluster-wide transaction when _SaveChanges()_ is called.
+
+* If the item's index (its version) on the server is different from the index you provide,  
+  _SaveChanges()_ will throw a `ClusterTransactionConcurrencyException`.
```
haludi: Maybe worth mentioning that the whole transaction will be rejected?

docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx
```diff
@@ -0,0 +1,365 @@
...
+    ```
+    </TabItem>
+    </Tabs>    
+
+---
+
+## Delete compare-exchange item using a store operation
+  
+* Use the `DeleteCompareExchangeValueOperation` [store operation](../client-api/operations/what-are-operations) to delete a compare-exchange item by its key and index, without opening a session.
+  This is ideal for stand-alone tasks where you need to perform a direct compare-exchange operation without involving document transactions.
```
haludi: Minor.  
The session is not only for handling documents but for handling multiple commands as one transaction.  
The underlying endpoint is handled by BatchHandler, and I think its name reflects its purpose.  
If you want to batch multiple compare-exchange commands as one transaction, it’s also a good use case.

docs/compare-exchange/content/_include-compare-exchange-items-csharp.mdx
```diff
@@ -0,0 +1,681 @@
...
+    // Access the included CmpXchg item:
+    // =================================
+
+    // Call 'GetCompareExchangeValue' to access the content of the included CmpXchg item.
+    // Pass the CmpXchg item KEY. This will NOT trigger another server call.
+    var item = session.Advanced.ClusterTransaction
+        .GetCompareExchangeValue<string>(company1.Supplier);
+
+    // You can assert that no further server calls were made
+    Debug.Assert(session.Advanced.NumberOfRequests == numberOfRequests);
```
haludi: Minor
It is weird for me to see in the documentation `Debug.Assert`, but maybe it is only me

docs/compare-exchange/content/_atomic-guards-csharp.mdx
```diff
@@ -0,0 +1,357 @@
...
+</TabItem>
+</Tabs>
+
+<Admonition type="note" title="">
+
+When _loading_ a document in a cluster-wide session, RavenDB attempts to retrieve the document from the document store:
+
+* **If the document is found**, it is loaded into the session,
+  and modifications will be saved successfully as long as no other session has modified the document in the meantime.
+  Specifically, if the document’s [change vector](../server/clustering/replication/change-vector) matches the one currently stored on the server,
```
haludi: Why do you mention the change vector here?
It is not relevant for cluster-wide transactions.
The change vector indeed contains some relevant information, but the only check is against the compare exchange index

haludi (commented):
docs/compare-exchange/content/_update-cmpxchg-item-csharp.mdx
docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx
docs/compare-exchange/content/_include-compare-exchange-items-csharp.mdx
docs/compare-exchange/content/_indexing-compare-exchange-values-csharp.mdx
docs/compare-exchange/content/_cmpxchg-item-expiration-csharp.mdx
docs/compare-exchange/content/_atomic-guards-csharp.mdx
docs/compare-exchange/configuration.mdx

docs/compare-exchange/content/_atomic-guards-csharp.mdx
Danielle9897: => 
```
RavenDB checks whether the Raft index of the atomic guard associated with the document matches the version tracked by the current session.
If another session has already updated the document (and incremented the atomic guard’s Raft index), the save will fail with a _ConcurrencyException_.
```

docs/compare-exchange/content/_create-cmpxchg-items-csharp.mdx
Danielle9897: Added a bullet:
```
* Note:  
   _PutCompareExchangeValueOperation_ is used for both **creating** and [modifying](../compare-exchange/update-cmpxchg-item) compare-exchange items.  
  The intent of the operation is determined by the index you pass:  
  * An index of `0` indicates a **create** operation.
  * A non-zero index indicates a **modify** operation.
```

docs/compare-exchange/content/_create-cmpxchg-items-csharp.mdx
Danielle9897: done

docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx
Danielle9897: =>
```
Compare-exchange tombstones that are eligible for deletion are removed periodically by an internal cleanup task.  
 ```

File: docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx (hunk 28 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx]: =>
```
This is suitable if you want to include compare-exchange deletions alongside other operations, such as putting or deleting documents and compare-exchange items, in a single transaction.
```

File: docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx (hunk 29 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx]: =>
```
* If the item's index (its version) on the server is different from the index you provide, _SaveChanges()_ will throw a `ClusterTransactionConcurrencyException`.
  This means the item was modified by another operation after it was loaded into the session, and the entire transaction will be rejected.  
```

File: docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx (hunk 30 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_delete-cmpxchg-items-csharp.mdx]: =>
```
This is ideal for stand-alone tasks that don't require batching multiple commands into a single transactional session. 
```

File: docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx (hunk 15 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx:72]: done

File: docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx (hunk 19 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx]: =>
```
Console.WriteLine($"Number of compare-exchange items retrieved: {items.Count}"); // Expecting 4
```

File: docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx (hunk 25 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx]: =>
```
Console.WriteLine($"Number of compare-exchange items with prefix 'employees': {items.Count}");
```

File: docs/compare-exchange/content/_overview-csharp.mdx (hunk 11 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_overview-csharp.mdx:282]: =>
```
// At this point, another process owns the resource.
// But if that process crashed and never released the resource, the reservation may have expired,
// so we can try to take the lock by overwriting the value using the current index.
```

File: docs/compare-exchange/content/_update-cmpxchg-item-csharp.mdx (hunk 26 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_update-cmpxchg-item-csharp.mdx]: =>
```
* The expected index (version) of the item, which must match the current version stored on the server.
```

File: docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx (hunk 16 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx:118]: done

File: docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx (hunk 17 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx:164]: done

File: docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx (hunk 18 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-item-csharp.mdx:205]: done

File: docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx (hunk 20 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx]: done

File: docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx (hunk 21 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx]: done

File: docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx (hunk 22 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx]: done

File: docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx (hunk 23 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx]: done

File: docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx (hunk 24 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_get-cmpxchg-items-csharp.mdx]: done

File: docs/compare-exchange/content/_include-compare-exchange-items-csharp.mdx (hunk 31 above)
Comment (Danielle9897) [reply docs/compare-exchange/content/_include-compare-exchange-items-csharp.mdx]: Modified to `Console.WriteLine`

File: docs/compare-exchange/content/_overview-csharp.mdx [hunk 33]
```diff
@@ -0,0 +1,302 @@
...
+```csharp
+private class SharedResource
+{
+    public DateTime? ReservedUntil { get; set; }
+}
+
+public void PrintWork() 
+{
+    // Try to get hold of the printer resource
+    long reservationIndex = LockResource(store, "Printer/First-Floor", TimeSpan.FromMinutes(20));
```
Danielle9897, line 237: =>
```
public void PrintWork() 
{
    // Try to get hold of the printer resource
    long reservationIndex = LockResource(store, "Printer/First-Floor", TimeSpan.FromMinutes(20));

    try
    {
        // Do some work for the duration that was set (TimeSpan.FromMinutes(20)).
        // Don't exceed the duration, otherwise the resource won't be available for someone else.
        
        // In a distributed system (unlike a multi-threaded app), a process may fail or exit unexpectedly 
        // before releasing the resource - never reaching the 'finally' block.
        
        // To prevent the resource from remaining locked indefinitely, 
        // we use a timeout to expire the reservation.
    }
    finally
    {
        ReleaseResource(store, "Printer/First-Floor", reservationIndex);
    }
}
```

haludi: >Don't exceed the duration, otherwise the resource won't be available for someone else.

If you exceed the duration, two components can use the same resource at the same time.
I would mention the motivation before I get into the implementation details, but it is only my opinion

PR #2404, opened 2026-04-22
RDoc-3786 Add strict canonical verification

poissoncorp:
### Issue link
[RDoc-3786](https://issues.hibernatingrhinos.com/issue/RDoc-3786) Rewrite canonical URLs between versions if there's an entry in redirects

### Additional description
(_didn't_ use ai to write it, so don't scan - read 😂)

This PR introduces a guarantee that every `<link rel="canonical">` won't cause 404 across all pages.
It introduces `DOCUSAURUS_STRICT_CANONICALS` env flag - when enabled, and if there's any invalid redirect or canonical mismatch, the build fails.

New features:
- Canonical redirects plugin - rewrites canonicals in every emitted HTML file using the redirect map from `scripts/redirects.json`, then verifies each rewritten canonical against the actual docs structure
- `npm run validate-redirects` CLI - comprehensive validation of `redirects.json` file. Checks if schema is correct and validates the data, runnable without a full build - added as a CI step
- Cloudfront function (`scripts/handle_redirects.js`) now handles the hops to pair it with plugin validation. Added parity tests and sprinkled with unit tests that check if the scripts works as expected and handles all corner cases before pushing to prod
- Legacy versions - unified the logic of skipping legacy (EOL) docs versions (exclude from sitemap, block robots, etc.) within `version-policy.js` which is a single source of truth. Whenever RavenDB version becomes legacy, simply adding version to this list cascades all necessary SEO ops.
- `templates-noindex-plugin` which hides `/templates/` from search engines

Redirects JSON fixes:
- The verify flow found 26 missing redirects for moved docs, that caused 404 canonicals - fixed by backfilling the redirects map
- Fixed one malformed pre-exsiting entry (`encrypted-backup`)

So, new infra consists of:
- 2 plugins - for rewriting canonicals and hiding templates, as mentioned above
- `handle_redirects` as it was before, now just tested before pushing to AWS :)
- `validate-redirects` CI tool
- `version-policy.js` as mentioned to control the legacy docs ops
- `generate-robots.js` renders `robots.txt` from template on every build, keeping legacy disallows in sync
- `split-sitemap` ported to TypeScript to make it testable

Sprinkled it with docs for agents, added `.svg`-s with UML diagrams on how the build flow works, and how redirect chains are resolved (the akward situation when we moved the document more than once between versions).

Once merged, **let's remember** to set strict canonicals env on CD. 

### Type of change

- [ ] Content - docs
- [ ] Content - cloud
- [ ] Content - guides
- [x] Bug fix
- [x] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs

### Changes in UX/UI
- [x] No changes in UX/UI


src/plugins/templates-noindex-plugin/index.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * Templates noindex plugin.
```
Lwiel: I think we can just use `unlisted` frontmatter option for template pages
https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-pages#unlisted

scripts/lib/version-policy.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Single source of truth for version policy. When a version ages out of support,
+ * move it from versions.json into LEGACY_VERSIONS here.
+ *
+ * CommonJS so Node scripts, compiled TS, and ES-module consumers can all import it.
+ */
+
+const CURRENT_VERSION = "7.2";
+
+const LEGACY_VERSIONS = [
```
Lwiel, line 10: Maybe setting `noIndex: true` in config is enough
https://docusaurus.io/docs/api/docusaurus-config#noIndex

poissoncorp: done - setting noIndex for legacy versions

src/plugins/templates-noindex-plugin/index.ts
poissoncorp: done - templates have unlisted flag

Lwiel (approved):

PR #2411, opened 2026-04-26
Update aggressive caching page

reebhub:
### Issue link
https://issues.hibernatingrhinos.com/issue/RDoc-3452
https://issues.hibernatingrhinos.com/issue/RDoc-2450
https://issues.hibernatingrhinos.com/issue/RDoc-3806

### Additional description
fix page texts & code, document missing features

### Type of change

- [x] Content - docs
- [ ] Content - cloud
- [ ] Content - guides
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs

### Changes in UX/UI
- [x] No changes in UX/UI

### DIFF
--- a/docs/client-api/how-to/content/_setup-aggressive-caching-csharp.mdx
+++ b/docs/client-api/how-to/content/_setup-aggressive-caching-csharp.mdx
@@ -2,35 +2,160 @@ import Admonition from '@theme/Admonition';
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 import CodeBlock from '@theme/CodeBlock';
+import Panel from '@site/src/components/Panel';
+import ContentFrame from '@site/src/components/ContentFrame';
 
-## Standard Cache Configuration
+<Admonition type="note" title="">
 
-The RavenDB client provides a caching mechanism out of the box. The default caching configuration is to cache all requests.
+* By default, the RavenDB client caches every HTTP response and revalidates it with the server using ETags and `304 Not Modified` on every read.
 
-The size of cache can be configured by changing [`MaxHttpCacheSize` convention](../../../client-api/configuration/conventions.mdx#maxhttpcachesize).
+* Aggressive caching skips that revalidation round trip:  
+   inside an aggressive caching scope, the client serves the cached response straight from local memory and never asks the server.
 
-The client utilizes the notion of the `304 Not Modified` server's response and will serve the data from the cache if available.
+* Use aggressive caching on read-heavy paths, on reference data that changes rarely, and on any code path where end-to-end latency matters more than absolute freshness.
 
-## Aggressive Mode
+* Cached entries are invalidated automatically through the [Changes API](../../../client-api/changes/what-is-changes-api.mdx).  
+   A short notification delay can briefly cause stale reads inside the scope.
 
-The aggressive caching feature goes even further. Enabling it means that the client doesn't need to ask the server. It will simply return the response directly from a local cache without any usage of `304 Not Modified` status. 
-Results will be returned very fast. 
+* Some requests bypass aggressive caching by design, including queries with `WaitForNonStaleResults`, queries with `NoCaching()`, and internal cluster traffic.
 
-Here's how it works: The client subscribes to server notifications using the [Changes API](../../changes/what-is-changes-api.mdx). By taking advantage of them, he is able to invalidate cached documents when they are changed.
-The client knows when it can serve the response from the cache, and when it has to send the request to get the up-to-date result. 
+---
 
-<Admonition type="warning" title="Important" id="important" href="#important">
-Despite the fact that the aggressive cache uses the notifications to invalidate the cache, it is still possible to get stale data because of the time needed to receive the notification from the server.
+* In this article:  
+   * [Caching in RavenDB](../../../client-api/how-to/setup-aggressive-caching.mdx#caching-in-ravendb)  
+      * [Standard HTTP caching](../../../client-api/how-to/setup-aggressive-caching.mdx#standard-http-caching)  
+      * [Aggressive caching](../../../client-api/how-to/setup-aggressive-caching.mdx#aggressive-caching)  
+      * [How invalidation works](../../../client-api/how-to/setup-aggressive-caching.mdx#how-invalidation-works)  
+   * [Enable aggressive caching](../../../client-api/how-to/setup-aggressive-caching.mdx#enable-aggressive-caching)  
+      * [Enable for a scope](../../../client-api/how-to/setup-aggressive-caching.mdx#enable-for-a-scope)  
+      * [Use convention defaults](../../../client-api/how-to/setup-aggressive-caching.mdx#use-convention-defaults)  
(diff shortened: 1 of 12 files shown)


poissoncorp:
The Python SDK has recently introduced aggressive caching (https://github.com/ravendb/ravendb-python-client/releases/tag/7.2.1). Not sure if documenting it should belong to this PR, though. Just fyi :)

reebhub:
@poissoncorp https://issues.hibernatingrhinos.com/issue/RDoc-3809/Python-create-a-Python-version-for-the-aggressive-caching-page

ppekrol (commented):
## Review of #2411 — aggressive caching docs

I validated the page against the v6.2 client implementation. The structural rewrite is a real improvement — the new TOC, the panel layout, the "Common mistakes" section, and the table that contrasts `AggressivelyCache()` / `AggressivelyCacheFor(d)` / `AggressivelyCacheFor(d, m)` all land well. Most factual claims line up with the code. The notes below are the things I'd want fixed (or at least considered) before this merges.

---

### A. Factual issues to fix

**1. "How invalidation works" describes per-entry eviction; the client actually invalidates globally.**

The page currently says:

> The client receives the notification and removes the matching entries from the local cache. The next read of any of those entries fetches a fresh copy from the server, after which subsequent reads are served from the cache again until the next change notification arrives.

In the v6.2 client this is not how it works. `EvictItemsFromCacheBasedOnChanges` does not look at *which* document or index changed — on any qualifying change it just bumps a single counter:

```csharp
// EvictItemsFromCacheBasedOnChanges.cs
public void OnNext(DocumentChange change)
{
    if (AggressiveCacheChange.ShouldUpdateAggressiveCache(change))
        Interlocked.Increment(ref _requestExecutor.Cache.Generation);
}
```

Every cached item then reports `MightHaveBeenModified == true` (it compares its own generation to the cache's), so the next read of *any* aggressively-cached item — not just the one that changed — bypasses the aggressive shortcut and revalidates over the wire. Items that come back `304 Not Modified` get their generation re-stamped one at a time; items that aren't read again stay marked.

For a developer this matters because the observable behavior after a single `Put` is "every aggressively-cached document I read next does a 304 round-trip", not "only the changed document". Suggested rewrite:

> When a document or index is modified on the server, the server publishes a change notification on that subscription. The client treats that as a signal that the local cache may be stale and marks every aggressively-cached entry for revalidation. The next read of each entry sends a single conditional request (`If-None-Match`) and is served from the cache as soon as the server confirms the entry is unchanged with `304 Not Modified`. Entries that are not read again stay marked until they are.

(That phrasing also makes it natural to add the dropped-connection note in A4 below.)

**2. Lazy row in "Requests that are never aggressively cached" is misleading.**

> Lazy operations are bundled into a single batched request and each one keeps its own cache rules, so a batch can have some operations served from the cache and others sent through.

For the *aggressive* shortcut path this is wrong. `MultiGetCommand.MaybeReadAllFromCache` is all-or-nothing:

```csharp
if (readAllFromCache && (trackChanges && cachedItem.MightHaveBeenModified
    || cachedItem.Age > options.Duration
    || command.CanCacheAggressively == false))
    readAllFromCache = false;
```

If any single item in the batch can't be served aggressively, the whole batch goes over the wire. The "some operations served from the cache and others sent through" behavior the doc describes is true on the response path of the *standard* 304 cache (server returns 304 per inner request), but it's not how aggressive caching behaves. Suggested rewrite for that cell:

> Lazy operations are bundled into a single MultiGet request. Aggressive caching can only short-circuit the batch when *every* operation in it is aggressively cacheable; if any one operation has `WaitForNonStaleResults` or `NoCaching()` set, the entire batch is sent to the server (individual operations can still be served from the standard 304 cache on the response path). See [Perform queries lazily](../../../client-api/session/querying/how-to-perform-queries-lazily.mdx).

**3. "Use this mode when the Changes API is unavailable, for example in some embedded scenarios."**

I couldn't find anything in `Raven.Embedded` that disables the Changes API, and an embedded server exposes the WebSocket endpoint normally. The legitimate reason for `DoNotTrackChanges` is the next sentence (skip the bookkeeping cost of the Changes connection, accept stale reads). I'd drop the embedded example unless there's a real scenario behind it, and lean on the second sentence:

> Use this mode when you explicitly accept stale reads in exchange for not maintaining a Changes API connection — for example, when the duration is short enough that staleness is bounded by it, or when the client is running in an environment where the Changes WebSocket is undesirable.

**4. Missing: what happens when the Changes connection drops.**

The client invalidates the cache as a safety on disconnect:

```csharp
// EvictItemsFromCacheBasedOnChanges.cs
public void OnError(Exception error)
{
    // any error means that the changes connection was disconnected, and we must invalidate the cache
    Interlocked.Increment(ref _requestExecutor.Cache.Generation);
}
```

This is exactly the question a developer running aggressive caching in production will ask ("if the websocket dies will I serve stale data forever?"). One sentence in "How invalidation works" closes that loop.

---

### B. Approachability — suggestions, not blockers

These are things I'd want as a developer reading this page cold. None of them are wrong; they're gaps.

**1. All body examples are sync.** The narrative sections use `documentStore.OpenSession()` and `session.Load<T>()`. A reader on an async pipeline has to scroll to the Syntax section to discover that `AggressivelyCacheForAsync` exists. I'd add one async example near "Enable for a scope":

```csharp
await using (await documentStore.AggressivelyCacheForAsync(TimeSpan.FromMinutes(5)))
{
    using (var session = documentStore.OpenAsyncSession())
    {
        var order = await session.LoadAsync<Order>("orders/1");
    }
}
```

**2. No guidance on choosing `Duration`.** The page tells you what `Duration` *is* (max age) but not how to pick a value. A short paragraph in "Enable for a scope" along these lines would help:

> The right `Duration` depends on how stale the calling code can tolerate the data being if the change notification is delayed or missed. Reference data that changes rarely is a good fit for hours or days; hot read paths where you still want some safety net work well at minutes; for code paths that must always see the latest committed value, don't use aggressive caching at all and rely on the standard 304 cache instead.

**3. The `DisableAggressiveCaching` example is a bit contrived.** Loading `orders/1` aggressively, then loading `orders/1` again with `DisableAggressiveCaching()` in the same `using` block reads as a toy example. The real use case is "this code is running inside an aggressive scope someone else opened, and for this one operation I need fresh data". Worth a one-line lead-in to that effect.

**4. Common mistakes — small wording tweak.** The "Forgetting to dispose" bullet says the scope "leaks into whatever code runs next on the same flow". For someone who hasn't internalized that this is `AsyncLocal`-based, "the same flow" is fuzzy. Consider:

> Without a `using` block, or an explicit call to `Dispose()`, the scope stays attached to the current execution context and is inherited by anything that runs from the same async chain — including subsequent requests in long-lived hosts.

---

### Summary

* **A1, A2, A3, A4** are factual or load-bearing. I'd block on those.
* **B1–B4** are polish. Take or leave.

The bones of this rewrite are good — these are mostly tightening, not redesigning.

reebhub:
@ppekrol done

ppekrol (approved):

PR #2433, opened 2026-05-10
RDoc-3765 RQL supports now() & today()

Danielle9897:
### Issue link
https://issues.hibernatingrhinos.com/issue/RDoc-3765/RQL-supports-now-today

### Additional description
- [x] Update the RQL article
- [x] Create a new article under **Querying > Filtering Query Results > Filter by Date and Time**
    - [x] Filter by an explicit date (single date, date-range, between)
    - [x] Filter using now()
    - [x] Filter using now() with offset (future, past, offset format)
    - [x] Filter using today()
    - [x] Combining now() and today()
    - [x] Restrictions
    - [x] Result caching with time functions
    - [x] Syntax

### Type of change
- [x] Content - docs
- [ ] Content - cloud
- [ ] Content - guides
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs
- [x] No changes in docs URLs

### Changes in UX/UI
- [x] No changes in UX/UI

### DIFF
--- a/docs/querying/filtering-query-results/content/_filter-by-date-and-time-csharp.mdx
+++ b/docs/querying/filtering-query-results/content/_filter-by-date-and-time-csharp.mdx
@@ -90,6 +90,8 @@ List<Employee> employees = await asyncSession.Advanced
 </TabItem>
 <TabItem value="RawQuery" label="RawQuery">
 ```csharp
+var cutoff = new DateTime(1995, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+    
 List<Employee> employees = session.Advanced
     .RawQuery<Employee>("from Employees where HiredAt >= $cutoff")
     .AddParameter("cutoff", new DateTime(1995, 1, 1, 0, 0, 0, DateTimeKind.Utc))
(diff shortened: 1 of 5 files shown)


docs/querying/filtering-query-results/content/_filter-by-date-and-time-csharp.mdx
```diff
@@ -0,0 +1,964 @@
...
+    </TabItem>
+    </Tabs>
+
+</Panel>
+
+<Panel heading="Filter using now() with offset">
+
+* `now()` can accept an optional positive or negative offset string, such as `+7d`, `-3d`, or `+1d5h`.  
+
+* When an offset is provided, RavenDB first **shifts the current time** by the specified amount
```
ppekrol: maybe add some flag that this is Important, bold? exclamation mark? upper case? not sure

docs/querying/filtering-query-results/content/_filter-by-date-and-time-csharp.mdx
```diff
@@ -0,0 +1,964 @@
...
+    <TabItem>       
+    ```sql     
+    // Not allowed: now() / today() are not supported in subscription RQL   
+    from Employees
+    where HiredAt <= now()
+    ```
+    </TabItem>
+
+  This restriction applies only to the `now()` and `today()` functions.  
+  These functions are blocked in exploration and subscription queries because their value depends on when the expression is evaluated, which can lead to caching correctness issues.
```
ppekrol: with subscriptions it is not a matter of caching, but yielding indeterministic results due to the fact that this is an ongoing task, so using now/today would dynamically (and drastically) change the results depending on the execution time

docs/querying/filtering-query-results/content/_filter-by-date-and-time-csharp.mdx
Danielle9897: done - added admonition & uppercase

docs/querying/filtering-query-results/content/_filter-by-date-and-time-csharp.mdx
Danielle9897: done

PR #2443, opened 2026-05-11
RDoc-3815 - update the replace-license page let's encrypt renewal section

reebhub:
### Issue link
[RDoc-3815](https://issues.hibernatingrhinos.com/issue/RDoc-3815) Document Let's encrypt cert renewal change

### Additional description
Update the replace-license page let's encrypt renewal section: contacting support is no longer needed to renew license.

### Type of change

- [x] Content - docs
- [ ] Content - cloud
- [ ] Content - guides
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs

### Changes in UX/UI
- [x] No changes in UX/UI

### DIFF
--- a/docs/licensing/replace-license.mdx
+++ b/docs/licensing/replace-license.mdx
@@ -28,7 +28,6 @@ import ContentFrame from "@site/src/components/ContentFrame";
 * In this article:
     * [Replace your license using a configuration key](../licensing/replace-license.mdx#replace-your-license-using-a-configuration-key)  
     * [Replace your license via Studio](../licensing/replace-license.mdx#replace-your-license-via-studio)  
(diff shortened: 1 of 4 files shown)


haludi (approved):
I don't see any mistakes in the text, but I found the current documentation a little bit weird.
It made sense to mention a certificate renewal of a replaced license server when it had an action the user needed to do.
Now, from the user's perspective, it is the same as renewing a certificate regardless of whether the license was replaced or not.
It can be relevant only for a user who read the old documentation.
I would remove that part entirely or at least pre-write that "unlike old RavenDB behavior currently..." to give the relevant context and avoid making the text more confusing than helpful.
But this comment is minor, and the current changes are totally acceptable.

reebhub:
ok, removed the "Let's Encrypt certificate renewal after a license replacement" section entirely

PR #2494, opened 2026-06-11
RDoc-3902 [Per-feature reorganization] Use logical operators

Danielle9897:
### Issue link
https://issues.hibernatingrhinos.com/issue/RDoc-3902/Per-feature-reorganization-Use-logical-operators

### Additional description
* Create new article: "Use logical operators" (C#)
   Cover: AND, OR, NOT, Negate a group, Mixed conditions, Operator precedence and Defaults
* This new article will replace the existing article: "How to use NOT operator"
* Create the Python, Node.js, PHP, Java articles

### Type of change
- [x] Content - docs
- [ ] Content - cloud
- [ ] Content - guides
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs
- [ ] No changes in docs URLs
- [x] Articles are restructured, URLs will change, mapping is required (update `/scripts/redirects.json` file, set `Documents Moved` PR label)

### Changes in UX/UI
- [x] No changes in UX/UI


scripts/redirects.json
```diff
@@ -962,5 +962,12 @@
             "targetUrl": "/querying/filtering-query-results/filter-with-lucene-syntax",
             "minimumVersion": "7.2"
         }
+    },
+    {
+        "key": "/client-api/session/querying/document-query/how-to-use-not-operator",
+        "value": {
+            "targetUrl": "/querying/filtering-query-results/use-logical-operators",
```
Lwiel, line 969: Maybe we should redirect to the relevant anchor? Not 100% sure if our redirects support it though

Danielle9897: Do you know if that will work ?
Either way -  imho - redirecting to the new article page is still good enough. 

The new page is now the main article for all logical operators.
It includes a dedicated NOT section, which is visible in the TOC.
So the old URL still lands users in the right topic area, and from there the NOT section is easy to find.

Lwiel: We decided to use link without anchor

Lwiel (approved):

PR #2525, opened 2026-06-19
RDoc-3857 Sitemap should contain only canonical URLs

poissoncorp:
Exclude non-current doc versions (7.1, 6.2, legacy) from the sitemap so it lists only the current canonical version's URLs. Stable `sitemap-docs.xml` naming and splitter cleanup are deferred to [RDoc-3955](https://issues.hibernatingrhinos.com/issue/RDoc-3955/Current-only-docs-sub-sitemap-with-stable-name).

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -115,7 +115,7 @@ const config: Config = {
                     lastmod: "date",
                     changefreq: null,
                     priority: null,
-                    ignorePatterns: LEGACY_VERSIONS.map((v) => `/${v}/**`),
+                    ignorePatterns: [...ACTIVE_VERSIONS, ...LEGACY_VERSIONS].map((v) => `/${v}/**`),
                 },
                 googleTagManager: {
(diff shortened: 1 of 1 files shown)


Lwiel:
Does it mean we can have a single `sitemap-docs.xml` file instead of `sitemap-docs-x.y.xml` where `x.y` is latest version?

poissoncorp:
@Lwiel 
> Does it mean we can have a single `sitemap-docs.xml` file instead of `sitemap-docs-x.y.xml` where `x.y` is the latest version?

Yes, but we should wait with that for the next release, so we don't cause SEO lag & reindexing just for tidiness. I've thought about it for a while and reduced this PR to minimal changes while the rest (simplification) will be done here - https://issues.hibernatingrhinos.com/issue/RDoc-3955/Current-only-docs-sub-sitemap-with-stable-name

Lwiel (approved):

PR #2561, opened 2026-07-16
RDoc-3454 Persist selected language in URL

poissoncorp:
### Issue link
[RDoc-3583](https://issues.hibernatingrhinos.com/issue/RDoc-3583) Persist selected language state in the URL
[RDoc-3454](https://issues.hibernatingrhinos.com/issue/RDoc-3454) Language switcher should open the article in a new tab on Ctrl+Click

### Additional description
Make the "?lang=" query parameter the per-tab source of truth for the selected API language, so a page can be shared/linked in a specific language and two tabs can show different languages side-by-side. On Ctrl+Click, a new tab appears with the clicked language.
 
### Type of change

- [ ] Content - docs
- [ ] Content - cloud
- [ ] Content - guides
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs

### Changes in UX/UI
- [x] No changes in UX/UI

### DIFF
--- a/src/components/LanguageStore.tsx
+++ b/src/components/LanguageStore.tsx
@@ -1,18 +1,41 @@
-﻿import { useSyncExternalStore, useCallback } from "react";
+import { useSyncExternalStore, useCallback } from "react";
+import { useHistory } from "@docusaurus/router";
+import { languageConfig } from "./languageConfig";
 
 export type DocsLanguage = "csharp" | "java" | "python" | "php" | "nodejs";
 
-const DEFAULT_LANGUAGE: DocsLanguage = "csharp";
+export const DEFAULT_LANGUAGE: DocsLanguage = "csharp";
+export const LANGUAGE_QUERY_PARAM = "lang";
+
 const LANGUAGE_STORAGE_KEY = "docs-language";
 
-const getLanguageFromLocalStorage = (): DocsLanguage => {
-    return (window.localStorage.getItem(LANGUAGE_STORAGE_KEY) as DocsLanguage) || DEFAULT_LANGUAGE;
+// Same-tab only (not the native cross-tab "storage" event) so two tabs can show different languages side-by-side.
+const LANGUAGE_CHANGE_EVENT = "docs-language-change";
+
+export const isDocsLanguage = (value: string | null | undefined): value is DocsLanguage =>
+    !!value && languageConfig.some((lang) => lang.value === value);
+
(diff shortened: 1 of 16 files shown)


Lwiel (approved):

src/components/LanguageStore.tsx
```diff
@@ -1,31 +1,70 @@
...
+        window.removeEventListener("popstate", callback);
     };
 };
 
 export const useLanguage = (): {
     language: DocsLanguage;
     setLanguage: (newLanguage: DocsLanguage) => void;
 } => {
-    const language = useSyncExternalStore(subscribe, getLanguageFromLocalStorage, () => DEFAULT_LANGUAGE);
+    const history = useHistory();
```
kalczur: URL hash changes (TOC clicks) now re-render every language block

src/components/LanguageUrlSync.tsx
```diff
@@ -0,0 +1,40 @@
...
+    supportedLanguages?: DocsLanguage[];
+};
+
+// Renders nothing. Keeps "?lang=" in sync with the active language (re-stamped after navigation)
+// and persists an incoming "?lang=" so the choice carries across navigation and future sessions.
+export default function LanguageUrlSync({ supportedLanguages }: LanguageUrlSyncProps): null {
+    const { setLanguage } = useLanguage();
+    const location = useLocation();
+
+    useEffect(() => {
```
kalczur, line 20: Now you can remove `useEffect` from `LanguageSwitcher.tsx`

kalczur (changes requested):

kalczur:
@poissoncorp 
To consider:
A Node.js user who browses a C# only page silently gets their global preference reset to C#, and later pages that supports Node.js open in C#. It would be better to update global preference only on user intended click.

src/components/LanguageStore.tsx
poissoncorp: done

src/components/LanguageUrlSync.tsx
poissoncorp: done

poissoncorp:
> @poissoncorp To consider: A Node.js user who browses a C# only page silently gets their global preference reset to C#, and later pages that supports Node.js open in C#. It would be better to update global preference only on user intended click.

@kalczur done

kalczur (approved):

PR #2600, opened 2026-08-17
[Quill documentation] RDoc-4044 - added a Quill overview page

reebhub:
### Issue link
[RDoc-4044](https://issues.hibernatingrhinos.com/issue/RDoc-4044)

### Additional description
Since other Quill articles haven't been added yet the article is missing links that will be added later on.

### Type of change

- [x] Content - docs
- [ ] Content - cloud
- [ ] Content - Quill
- [ ] Content - guides
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs

### Changes in UX/UI
- [x] No changes in UX/UI


quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
+---
+title: "Quill: Overview"
+sidebar_label: Overview
+sidebar_position: 1
+description: "Quill is a service that mirrors selected tables of your existing database and lets your users ask questions about your data through an AI chat widget."
```
Scooletz:
```suggestion
description: "Quill is an AI data layer over your existing relational database. It mirrors the data you select requiring no migration, no schema changes, no queries hitting production databases."
```

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+* Databases like PostgreSQL, SQL Server, and MySQL lack AI capabilities that could greatly improve user experience.  
+  For example, a company that runs one of these databases may want to offer its website visitors a chat widget that answers
+  their questions using the company's live data.  
+  To accomplish this, the company would have to either build the chat widget interface from scratch, an effort that requires
+  expertise in AI, databases, and security, or migrate to a database that already offers such capabilities, a major
+  undertaking as well.
+
+  **Quill** is a ready-made alternative that spares you both efforts: it runs as a service in a Docker container beside your
+  database, reads your live data using access details you provide, and publishes AI interfaces (like the chat widget mentioned
```
Scooletz, line 23: I like the fact that a chat is referred as an example. This means that when we update Quill with new capabilities, this description does not get immediately stale.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+  above) that offer your users the AI experience you want to incorporate.  
+  Your existing infrastructure, including your data, workflows, and applications, remains unchanged.
+  
+* Quill reads from your database only to mirror part of your data: you select the tables it may draw answers from, and
+  Quill keeps a live copy of these tables in an internal database.  
+  Replies are drawn from this internal database, while your own database goes on serving your applications exactly as before.
```
Scooletz, line 29: Is it intentional, the line split here? Feels like something is wrong. Anyway.

```suggestion
* Quill imports data from your database only to mirror part of it you select. Data are imported initially to its internal database and kept in sync.
  Any interaction with Quill draws data from this internal database, while the original relational database goes on serving your applications exactly as before.
```

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+</Admonition>
+
+<Panel heading="What Quill is">
+
+Quill is a complete AI service for your PostgreSQL, SQL Server, or MySQL database, packaged as a single Docker container.
+
+Quill's Docker container includes everything Quill runs:
+
+* The machinery that mirrors your source database's tables
```
Scooletz, line 54: It's not mirroring as we can do more with the data. We can embed documents, link them, in the future we will use gen AI and embeddings. We need a better verb for this.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+* An internal database that holds the mirrored data
```
Scooletz, line 55: As above. Mirrored

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+* AI agents that talk with your users, query the mirrored data, and compose the replies
```
Scooletz, line 56: `mirrored`

Scooletz, line 56: An interesting framing, is it intentional? I'd phrase it as _Users talk with your AI agents_

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+#### Your applications
+
+* Your applications, like your online store or your in-house tools, keep reading from and writing to your SQL database as usual.
+* Quill takes no part in this traffic.
+
+#### Your SQL database
+
+* Your existing SQL database: PostgreSQL, SQL Server, or MySQL.
+* Quill reaches your database using a connection string you provide, and reads only the tables you select.
```
Scooletz, line 80: It feels that Quill does it all the time. There are two separate parts, the initial import/ingestion and the latter CDC that notices changes and imports only what is needed. Not sure what is the intent here, but I think we should explicitly split these two parts. Otherwise, it feels that Quill reads it all the time

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+* Your data is only read, never modified.
+
+#### Mirroring
```
Scooletz, line 83: Again, the verb. Not sure about it

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+* Quill begins mirroring your data by reading the selected tables from your SQL database in full, and copies their current
```
Scooletz: Shall this be prefixed with `During the initial setup`? This would make it safer and closer to the truth.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+  contents into its internal database.  
+  This initial operation may take some time when large tables are copied; if the operation is interrupted, Quill will
+  resume it where it stopped rather than start over.
+* To keep its internal database current, Quill uses your database's **change-recording** facility.
```
Scooletz:
```suggestion
* To keep its data fresh, Quill uses your database's **change-recording** facility.
```

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+  * PostgreSQL, SQL Server, and MySQL each provide a built-in facility that records every change made to the data.
```
Scooletz, line 90:
```suggestion
  * PostgreSQL, SQL Server, and MySQL each provide a facility that records every change made to the data.
```
Nitpicking maybe, but I'd omit built-in to make it about Quill, not things that are already there in the database.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+  * Quill continuously reads this record of changes, and applies each insert, update, and deletion made in the selected
+    tables to the internal database.
```
Scooletz, line 92: What is with the newlines here? Is it intentional?

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+    tables to the internal database.  
+    Preparing the change-recording facility for the selected tables is part of Quill's setup; if permissions are
+    missing, Quill will provide the exact script for your database admin to run.
+* The change-tracking technique described here is called **CDC (Change Data Capture)**.  
+  A reader of the recorded changes, like Quill, is called a **CDC sink**.  
+  The term "CDC" appears in Quill's dashboard as well, naming the views and settings related to mirroring.
+
+#### Internal database
+
+* The result is a **mirrored database**: a live copy of the selected tables, kept current as your data changes.
```
Scooletz: mirrored strikes back 😛

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+* Quill's internal database is a RavenDB document database, and Quill therefore stores the mirrored data not in tables but
+  as JSON documents.  
+  During setup, Quill suggests how the rows of each selected table become documents, and you can accept or adjust the
+  suggestion.  
+  For example, a customer's orders, read from your SQL database, can be stored in the **customer** document, or kept as
```
Scooletz: I'd use `order` and `order lines` which are more natural and easier to follow. I've never witnessed orders glued to customers.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+  separate **order** documents that link to the **customer** document.
+
+#### AI agent
+
+* Agents are AI components that Quill uses to handle user conversations.
+* You can provide your agents with **query tools** and **actions**, and the agents will apply both freely during
```
Scooletz: This feels like the setup/implementation overhead. Should we use passive here as agents can be provided with tools by the wizard / Quill itself, with no human intervention. We should emphases the ease of use and the way it's done for the user by Quill, with minimal action needed

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+* Agents draw their answers from Quill's internal database (mirroring your SQL database), so their replies to users
+  are based on your actual, current data records rather than on the LLM's general knowledge.
+
+</ContentFrame>
+
+<ContentFrame>
+
+### Answering questions
+
+Each exchange between a user and an agent starts at the chat widget and is answered from the mirrored database, a full
```
Scooletz: Should this be chat specific, or could we extend it to a generic notion of a channel? I think it is true for any channel

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+RavenDB database running inside the Docker container.
+
+* When a user types a question in the chat widget, the agent queries the mirrored database for the data the answer needs.
```
Scooletz:
```suggestion
* When a user types a question in the chat widget, the agent queries the Quill's database for the data the answer needs.
```

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+<ContentFrame>
+
+### One Quill, several apps
+
+* An **app** groups the working parts described above: a connection to one source database, the mirrored database
+  built from it, and the agents and channels that answer from this database.
+* A single Quill container can run several apps, each with its own source, mirrored database, agents, and channels
+  (e.g., one app for your store and another for your support system).
+* Quill's dashboard lists your apps side by side and is where you create and manage them.
+
```
Scooletz, line 171: Again, so much mirroring that it feels that we just mirror things.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+so its management dashboard and the [chat widget](overview.mdx#chat-widget) it runs are served over HTTPS from the
+moment Quill first starts, with no certificate handling on your side.
+
+* Quill reaches [your SQL database](overview.mdx#your-sql-database) using the access details you provide; your data
+  is only read, never modified.
+* Chat widget pages are the only public part of the service.
+* A chat widget page is reached through an **embed link** you place in your page.  
+  Each link expires, serves a limited number of exchanges, and can be revoked at any time.
+* Quill's management dashboard requires the API key you receive at sign-up.
+* Direct access to [Quill's mirrored database](overview.mdx#internal-database) requires a client certificate.
```
Scooletz:
```suggestion
* Direct access to [Quill's database](overview.mdx#internal-database) requires a client certificate.
```

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+* Quill reaches the [LLM provider](overview.mdx#llm) under your own account, using the access details your provider
```
Scooletz:
```suggestion
* Quill reaches the [LLM provider](overview.mdx#llm) using the access details your provider
```

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+  requires.
```
Scooletz: What's up with these newlines?

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+The stages below establish a **Quill deployment**: a running Quill container with a web address, a license, and a
+management dashboard of its own.
+
+The road from sign-up to a working AI interface:
+
+1. **Sign-up**  
+   *The deployment gets its identity.*  
+   On the sign-up page, you choose a name for the new deployment and provide the address of the machine that will
+   run Quill; the name becomes the deployment's web address.
```
Scooletz: > the name becomes the deployment's web address

I do not understand it. Is it the domain? What is the name that becomes the web address?
If it's about the name, the notion of _quill subdomain_ as in _your-sub-domain.quill.run_ can be used.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+   *The deployment gets its identity.*  
+   On the sign-up page, you choose a name for the new deployment and provide the address of the machine that will
+   run Quill; the name becomes the deployment's web address.  
+   The page returns a license key, an API key, and a ready-to-run install command.
+
+2. **First start**  
+   *Quill starts running on your machine.*  
+   The install command downloads Quill and starts its Docker container.  
+   During this first start, Quill activates itself using the license key, and then exposes the management dashboard
+   over HTTPS.
```
Scooletz, line 211: new line strikes back

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+3. **Database connection**  
+   *Quill connects to your SQL database and starts mirroring it.*
```
Scooletz, line 214: `mirror`

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+4. **LLM connection**  
+   *Quill gets access to an LLM.*  
+   You store the access details of your LLM provider (e.g., an OpenAI API key) in the dashboard, as an
+   **AI connection string**.  
+   Agents will use this connection string to connect to the LLM and phrase their replies.
+
+5. **Agent creation**  
+   *The deployment gains its first agent.*  
+   Quill examines the mirrored data and drafts a few agent configurations you can start from (e.g., a catalog
```
Scooletz: Should boast a bit more here? It's an AI that proposes AI, no human involved until the very last step.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+<Panel heading="Working with Quill using code">
+
+Developers can work with Quill using code as well as through the management dashboard.
+
+* **Managing Quill through its API**  
+  Your Quill deployment exposes an API; requests to this API are authenticated using the API key that sign-up
+  provides.  
+  The API covers Quill's management operations, like connecting a data source, creating agents and channels, and
+  generating embed links.  
+  e.g., your site can use the API to generate a personal embed link for each signed-in user.
```
Scooletz: This is put here just as an example, while it's a quite a feature on its own. Being able to configure channels so that they can be end-user-specific. Maybe this will be covered separately.

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+  A minimal API call, listing the deployment's apps:
+
+  ```bash
+  curl -H "X-Api-Key: <your-api-key>" https://api.<your-domain>/api/apps/
+  ```
+* **Reading the mirrored database directly**
```
Scooletz: Do we want to mention it? If yes, shall it be in the overview?

quill/overview.mdx
```diff
@@ -0,0 +1,303 @@
...
+  database.
+
+  Connecting with the C# client:
+
+  ```csharp
+  using var store = new DocumentStore
+  {
+      Urls = new[] { "https://db.<your-domain>" },
+      Database = "<app-database>",
+      Certificate = new X509Certificate2("quill.client.pfx")
```
Scooletz: This automatically raises a question where do I take this certificate from. I think this makes me even more opinionated in regards to splitting this into a separate advanced page.

Scooletz (changes requested):
Requires a few amendments.

quill/overview.mdx
reebhub: (note - this description is for search engines only, it doesn't appear on page.)
Took both points - "no migration, no schema changes" is in, and the description now says "relational database" with the three engines as examples. 
I kept "AI agents answer your users" rather than "AI data layer", since this text is only shown in search results, and people search for the concrete words, not for a term they won't find anywhere else in the docs.

-->
description: "Quill adds AI capabilities to your existing relational database (PostgreSQL, SQL Server, or MySQL): it mirrors the data you select, with no data migration and no schema changes, and lets AI agents answer your users from the mirrored data."

quill/overview.mdx
reebhub: right, the chat widget is presented everywhere on the page as one example of an AI interface, so new capabilities won't invalidate the text

quill/overview.mdx
reebhub: you're seeing line breaks made so the text will be more readable in an IDE, it just looks strange in github.
(would also be a much nicer experience for you to render the article via Docusaurus & see it the way our users will.)

regarding the wording: "keeps a live copy" carries the "initial import, then sync" fact, and the Mirroring section below details both phases in full. "Replies are drawn" is deliberate too, the top note tells the chat story, and each kind of access is then covered in its own section further down.

quill/overview.mdx
reebhub: after quite an extensive investigation **mirroring** is still in the lead. not perfect, but better than everything else. 
close contenders are, for example, **reflect** (good replacement until you start using it in sentences and get to “the reflected database”, “data reflection” etc), **capture** (came close, also for being the last word of CDC, but lacks the sense of continuity - “the captured database” for example sounds stiff), **projected** for already having a major RavenDB meaning, **reproduce** (another good candidate that falls because it has no sense of ongoing change), and various others that do make sense but fail when you take a closer look (tracked, echo, imitate, shadow, **sync** for being two-way, import, and so on).
so **mirroring** stays unless we have a specific better term, mainly because of how natural it is in reflecting the ongoing change tracking. We do state that the data is not copied as-is (when we talk about its transfer from table rows to JSON documents), and when GenAI enters the picture we will need to describe the internal database more in terms of a “derived database”. but this can be dealt with.

quill/overview.mdx
reebhub: see above

quill/overview.mdx
reebhub: see above

reebhub: seems more obvious when you look at it in its bullet list on the page: the bullet is part of the component list where every line leads with the component and what it does (the machinery that mirrors, the channels that carry...). flipping the subject to the users would break the structure and sound odd. The users' point of view gets its own places below, under "Your users" and "Answering questions".

quill/overview.mdx
reebhub: the split is there - it's the next component down. "Mirroring" is built on exactly those two phases: "Quill **begins** mirroring your data by reading the selected tables in full", and the next bullet - "**to keep its internal database current, Quill uses your database's change-recording facility**... continuously reads this record of changes, and applies each insert, update, and deletion". 
the bullet you commented on says that Quill reads only the selected tables and never changes your data, and this is true both during the initial import and during the ongoing change-reading.

quill/overview.mdx
reebhub: right, discussed above - not a perfect term, but a perfect one hasn't been found yet. "mirroring" covers the ongoing change tracking naturally, and requires extra explanations wherever the data is transformed. currently, only where we cover the transfer from table rows to JSON documents. in the future, when we explain how GenAI changes the imported data before it lands in the internal database.

quill/overview.mdx
reebhub: done

quill/overview.mdx
reebhub: right, it's the data that needs to remain current (sounds better than "fresh" in my ears), not the database.
-->
* To keep the data in its internal database current, Quill uses your SQL database's change-recording facility.

quill/overview.mdx
reebhub: this sub-bullet explains what the databases provide. "built-in" reassures readers who wonder whether they need to install anything that they don't, it's built-in.

quill/overview.mdx
reebhub: yup, explained above

quill/overview.mdx
reebhub: such a great word (o:

quill/overview.mdx
reebhub: done
-->
For example, an order's **lines**, read from your SQL database, can be stored in the **order** document, or kept as separate **order-line** documents that link to the **order** document.

quill/overview.mdx
reebhub: -->
* Agents carry two kinds of tools, and apply both freely during conversations: **query tools** and **actions**.  
  Quill drafts an agent's query tools for you; you can adjust them, add your own, and define the agent's actions.

quill/overview.mdx
reebhub: -->
Each exchange between a user and an agent starts at a channel, like the chat widget on your site, and is answered with data from the mirrored database, a full RavenDB database running inside the Docker container.

quill/overview.mdx
reebhub: done

quill/overview.mdx
reebhub: -->
* A single Quill container can run several apps, each with its own source, database, agents, and channels
  (e.g., one app for your store and another for your support system).

quill/overview.mdx
reebhub: --> Quill's internal database

quill/overview.mdx
reebhub: done

quill/overview.mdx
reebhub: improves the text's readability in the IDE
recommended: read the articles in the Docusaurus environment, not here

quill/overview.mdx
reebhub: the sentence is vague because there is currently an ambiguity between two domain names: the sign-up page says the deployment will be available at `<name>.quill.run`, but the setup package, the DNS records, the certificates, and the embed URLs all use `<name>.myquill.ai`.
I will change the text now to "the name becomes the deployment's subdomain: `<your-name>.myquill.ai`", let me know if it should be `<your-name>.quill.run` instead.

quill/overview.mdx
reebhub: all hail new lines (o:

quill/overview.mdx
reebhub: all the rest hail mirroring (o: (o:

quill/overview.mdx
reebhub: clarified and added examples
generally I feel we're giving gold, so all we need to do is show it clearly enough, no additional sales pitch required

quill/overview.mdx
reebhub: yes, of course, everything in the overview page is just a taste of the real deal, all the sections here will include links to their full pages as soon as they are ready. a page that explains how to embed widgets is on its way.

quill/overview.mdx
reebhub: I think so. It's a supported capability (the deployment provides a client certificate for it), the security panel mentions the certificate requirement, and imho it's good publicity for RavenDB.
If you think it shouldn't be here, do tell and I'll remove it.

quill/overview.mdx
reebhub: -->
Your applications can connect to this database using RavenDB client libraries and a client certificate that you generate in the dashboard's **Certificates** view.

and as mentioned above, this is an overview-level taste - we will link it to the relevant locations when the pages are written.

reebhub (commented):
thanks for the thorough review

quill/overview.mdx
Scooletz: Process-wise, if it does not hurt much, I think we should make it readable in GH as well.

My point is that if you make it "just a copy", the premise is much weaker. I like the current shape of it much better.

quill/overview.mdx
Scooletz: Thank you for providing such detailed reasoning about the wording. I don't have a better candidate. My worry is that a mirror or a copy is a bit downplaying, we do so much more. Still, as I have no better proposal for it, let's **mirror** it.

quill/overview.mdx
Scooletz: "data fresh" is more opaque on purpose. "data current" is also good. My intention is that whenever we can make the Quill documentation a bit product-ish and less engineering, it's a good direction.

quill/overview.mdx
Scooletz: Ah, I see. I did not read it this way. The question is will the readers read it the way author intended? 👀

reebhub: very true. it's much clearer on the actual page I think, because the bulleting is shaped to emphasize  exactly that.

Scooletz (approved):
Good to go. All remarks addressed. The last outstanding `mirror`. We agreed with @reebhub that it's the best option out there. If a better is found in the future, we can revisit.

PR #2393, opened 2026-04-09
RDoc-3556 better docs for automatic storage extension

gregolsky:
### Issue link
https://issues.hibernatingrhinos.com/issue/RDoc-3556

### Additional description
...Include details of the change made, paste screenshots if necessary. Anything that may be useful for the reviewers...

### Type of change

- [ ] Content - docs
- [x] Content - cloud
- [ ] Content - guides
- [ ] Content - start pages/other
- [ ] New docs feature (consider updating `/templates` or readme) 
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs
- [ ] Articles are restructured, URLs will change, mapping is required (update `/scripts/redirects.json` file, set `Documents Moved` PR label)

### Changes in UX/UI
- [x] No changes in UX/UI
- [ ] Changes in UX/UI (include screenshots and description)

PR #2640, opened 2026-08-31
RDoc-4094 Cloud docs update for the September 2026 deployment

PFYasu:
Documents the user-facing changes in the September deployment window.
Ticket: https://issues.ravendb.net/issue/RDoc-4094

## Alerts

- New **Alerts** section in *Maintenance & Troubleshooting*: daily summary vs. custom alerts, subscribers, the per-product limits (25 rules, 20 recipients, one daily summary) and the four cases where rules are removed automatically — including a last-recipient removal deleting the whole rule.
- Trigger events point at the existing **Labels** table rather than restating them, since that table already carries the thresholds.
- Short generic **View the Alerts** section on the Products page, following the Metrics / Cluster Health pattern (generic in the Portal pages, detail outside).

## Billing & Costs

- **Current billing period costs** donut, the **Unbilled** total and the *Cost Details* entry points.
- **Spending**: the cost-category filter (all 18 categories), the year-picker behaviour and the "not calculated yet" states.
- New **Cost details** section for `/billing/cost-details`, which was previously undocumented, including the prediction caveat on unbilled periods and the marketplace notices.

## Portal navigation

Rewrote the Portal overview for the collapsible sidebar: the seven role-gated views, the **Help center** menu (Support, Documentation, Community, Service status) and the profile entry. Support and Service status are no longer top-level entries. Added the missing *Users & Access* entry, and noted that the account switcher identifies accounts by **domain name**.

## Support

**Request support for this instance** from the product view (paid support only, pre-selects the product), plus the support resources panel on the product view.

## Login

Google, GitHub **and Microsoft** sign-in, with each provider's email-verification requirement. Microsoft work/school (Entra ID) accounts additionally need the tenant to attest domain ownership via the `xms_edov` optional claim — worth calling out, because a tenant that does not emit it fails with "could not be verified" even though the address is correct. Also lists the actual sign-in failure messages.

## Provisioning

Noted that the **Account** and **Billing** wizard stages are skipped when the account already has those details, and are never shown for marketplace accounts — so the documented five-stage flow is the first-product case.

---

# Second pass — validated against `hr/test`

Every claim below was checked against the cloud portal source rather than inferred, and the whole set was re-validated after the September merge landed on `hr/test`.

## Autoscaling — rewritten (RDBCL-3989)

RDBCL-3989 replaced the flat CPU thresholds with three independent metrics: **load average**, **available memory** and **IO wait**, each with its own low/high limit and duration. The old section described CPU only, and 4 of its 10 table rows no longer exist in the UI.

- Separate **Upscaling** and **Downscaling** tables, each row carrying the UI's enforced min/max from `autoscalingThresholdFields.ts`.
- New **Autoscaling range** subsection for lowest/highest type.
- Admonition on the asymmetry the code comments call out: downscaling watches the *low* load average and *low* IO wait but the *high* available memory, because available memory runs opposite to utilisation.
- New **Serverless Autoscaling** subsection — the one `ProductFeatureName` value that had no coverage. Always enabled (`canDisableFeature={false}`), and its modal exposes only the instance range; the thresholds are managed for you.

## Automatic Storage Extension — corrected

- The threshold range is **70–95**, not 0–100. The `<input>` carries `min=0 max=100`, but both `UpdateThresholdSchema.ts` (`Yup.number().min(70).max(95)`) and `ProductValidator.cs` enforce 70–95. Entering 50 was rejected with no explanation in the docs.
- Two gating rules were undocumented (`ModelExtensions.cs`): the product must be **Active**, and the threshold **cannot be changed at all on yearly contracts**.
- The setting is **disk** usage, not memory — the old wording came from a screenshot predating the `Memory Used Threshold` → `Disk Space Used Threshold` rename.

## Products tab — restructured

- **Security** and **Allowed IPs** split into separate sections; they are separate cards and are not adjacent in the UI.
- Documented the **Certificate expiration date** card (RDBCL-4473) and the renewal schedule behind it: certificates are issued for 5 years, a renewal email goes out at 60 days, and automatic regeneration happens at 30 days (`Constants.Certificate`, `ProvisionProductCommandHandler`).
- **Instance Access** rewritten — `ProductEditAccessSection` is dead code on `hr/test`; Open Studio moved to the product-details top bar and becomes a **per-node dropdown** on multi-node products, so "opens the Studio of node A" was wrong for clusters.
- New **Subscription & support** section (RDBCL-5326): subscription details, the **Update available** release-channel pill, the marketplace guard, and the `Support`-role gating on the badge and CTA.
- Section order now follows the UI, and figure numbering was reconciled.

## Support — response times and the skipped step

- The response-time admonition stated two conditions; the code has three. **Severity is the dominant gate** and was entirely undocumented: *Within 2 hours* additionally requires `severity === "Critical"`, *Next business day* requires Normal or Critical. A paying customer filing a Minor issue finds both greyed out. Replaced with a per-option condition table.
- The tier restriction applies only to product-related questions — a general question is never tier-blocked.
- Both step lists were missing a **mandatory** step: the **Resources that may help you** panel gates the rest of the form via `isFamiliarWithResources`, so the Expected Response Time section does not render until you click through it. Lists are now 7 and 10 steps, using the UI's own field labels.

## Backup

The mandatory backup is a **Server-Wide Backup** (`BackupHelper` takes a `ServerWideBackupConfiguration`), so it covers every database rather than being per-database. The navigation said "Manage Ongoing Tasks"; the view is **Ongoing Tasks**.

## Link and alt-text hygiene

- **10 broken in-page anchors fixed** across `cloud/` — verified old-broken / new-valid against the rendered HTML, not guessed slugs.
- **152 image tags normalised.** `"Figure N - X"` put the literal quotes *inside* the alt attribute (`alt="&quot;…&quot;"`), which screen readers announce. The `Figure N` prefixes were invisible (`LazyImage` renders no caption) and referenced by no prose, while costing renumbering on every insertion. Now `X`. Also repaired a malformed tag in `cloud-aws-marketplace.mdx` that was missing its closing quote.

---

# Third pass — folding in RDoc-3556

[RDoc-3556](https://issues.hibernatingrhinos.com/issue/RDoc-3556) was closed as *Obsolete* together with its
PR **#2393**, which never merged. Only three of its claims (the 70–95 range, the yearly-contract block, and the
`Memory` → `Disk` correction) were already covered by the second pass above; the rest was not in the repo at all.
Its content is folded in here, **re-validated against `hr/test` (`83c55efe7`)** rather than carried over as written —
one claim did not survive and several gaps in it were filled.

## Automatic Storage Extension — the rest of the section

- **Availability.** Development and Production tiers, excluding **Dev10** and **PN**-type instances
  (`ProductFeaturesHelper.IsFeatureAvailable`). Also noted that an ineligible product has no feature row at all —
  `GetProductFeaturesDto` skips the row when the feature is absent from the PPC (RDBCL-5096), so "I can't find the
  setting" and "the setting is off" look identical in the UI.
- **Enabled by default at 90%, and free.** Verified both halves: RDBCL-4266 added it enabled by default and the
  `20260317` db-patch backfilled existing products at 90; `GetFeaturePercentage` returns **0** for
  `AutomaticStorageExtension`, so the feature carries no charge and only the larger disk is billed. #2393 asserted
  "free of charge" without this, and there *is* an `AutomaticStorageExtensionFeature` cost line type, which makes
  the claim look wrong until you follow the percentage.
- **How it works.** The node reports usage to `/api/v2/node/storage/auto-extend`; the disk then moves to the
  **next size the instance type offers**, not to the size that would merely clear the threshold — worth stating,
  because the jump is often much larger than the overflow. The success email carries the old, new and maximum
  size (`AutoStorageExtend.html`).
- **Azure Premium disks** also gain the new size's **IOPS and throughput** when the target performance tier is
  better (`ExtendProductStorageVolumeHelper`). Undocumented anywhere, and it changes the bill.
- **When storage is not extended.** #2393 listed three cases from a guess at the enum; `StorageExtensionForbiddenReason`
  has five, and they do not map one-to-one onto emails. Documented the four that notify you — instance-type maximum,
  root-disk limit, yearly contract, Free tier — and left out `StorageSizeLocked` (PN instances), which sends nothing
  and cannot be reached anyway, since PN products are not eligible. Also corrected the framing: the root-disk limit
  (**1,000 GB** AWS/GCP, **1,024 GB** Azure) applies **only to root-only disk layouts**, not to every product.
- **One warning per day.** `EmailNotificationPolicy` re-registers each of these notifications at most once per
  24 hours per product.

## Storage modification rate — new, and not from #2393

`Constants.StorageChangeLimitExpirationMinutes` enforces **6 h** on AWS and **12 h** on Azure Premium between two
storage changes. It gates `CanChangeStorage` and `CanExtendStorageAutomatically` alike, so it silently delays an
automatic extension: a disk filling up fast can sit above its threshold for hours. Documented in the feature section
and, since it applies to manual resizes too, on **Scaling → Change Storage**.

## Also

- Cross-link from **Scaling → Change Storage** to the feature (#2393 added this too).
- Explicit `{#storage-extension-configuration}` slug on the new *Configuration* heading — the page already has three
  *Configuration* sections, so it would otherwise have become `configuration-3`.
- Anchors verified against the rendered HTML, as in the second pass. `npm run build:current` clean.

---

## Before merge

- [ ] **`cloud/portal/assets/portal-product-alerts.png` is a placeholder.** Component-level capture with no portal chrome, so it does not match the framing of the neighbouring Metrics / Cluster Health figures. Needs a real-portal shot (sidebar + product header + Alerts tab active).
- [ ] Sections still without figures: **Cost details**, the Alerts subsections in *Maintenance & Troubleshooting*, and **Serverless Autoscaling**.
- [ ] Stale figures: `portal-billing-costs-tab-current-billing-period.png` still shows the 3% marketplace commission notice removed by RDBCL-4746; `portal-billing-costs-tab-billing-history.png` predates two column reworks (RDBCL-4739, RDBCL-4751). SNMPv3 Monitoring credentials has no figure while its sibling does.
- [ ] **Terminology.** The prose describes a sidebar and calls them "views", while the eight portal page titles and filenames still say "Tab". Renaming needs `scripts/redirects.json` entries, so it is out of scope here.

## ⚠️ Deploy ordering

Two features documented here are on `hr/test` but **not** on `hr/prod`:

| Feature | `hr/test` | `hr/prod` |
|---|---|---|
| Autoscaling metrics (RDBCL-3989) | yes | **no** |
| "Sign in with Microsoft" (RDBCL-5283) | yes | **no** |

Correct for the September deployment; **wrong if this merges before that ships.** Please check the deploy order before merging.

## Noted while working

`#manage-product-the-security-tab` exists on no page, so three guides were landing readers at the top of the Products page. **Fixed here** — and not with the anchor first suggested: the link text is *"For more details on adding a new product"*, so the target is `#provisioning-a-new-product`, not the security section. Each link keeps its existing absolute or relative form; only the fragment changed.

- `guides/connecting-c-application-to-ravendb-cloud.mdx`
- `guides/connecting-node-js-application-to-ravendb-cloud.mdx`
- `guides/how-to-setup-ravendb-with-asp-net-core-8-application.mdx`

(The previously-noted `#cloud-account-portal-products` from `cloud/cloud-migration.mdx` is fixed here too.)

Docusaurus validates page links but not anchors, which is why the build stays green either way — every anchor in this PR was checked against the rendered HTML instead.

Still unreferenced, and orphaned before this work: `cloud/assets/portal-product-details-storage-upgrade-to-premium-v2.png`, a duplicate of the `portal/assets` copy.

Verified with `npm run build:current` — clean, no warnings, no broken links.

---

## Asset naming convention

Filenames now read `<surface>-<view>-<region>[-<state>]`, so the name is the path you navigate to re-shoot the figure. **75 files renamed**, all as `git mv`.

- **Six pointed at the wrong place**: `portal-tabs` (it is the sidebar, not tabs), `portal-support-subscriptions-tab` (it is on the product view), `provider-region-tier-selection` (the Plan stage, completing the five-stage wizard set), plus the two emails, now under an `email-` prefix.
- **The 38 positionally-numbered `<page>-NNN-<desc>` names are retired.** That scheme had already collided twice — two `pricing-001` files with no `pricing-002`, and two `tiers-and-instances-0034` — which is the same defect as the `Figure N` alt-text prefixes this PR already removed. Redistributed by real capture surface: `portal-`, `studio-`, `os-windows-`, `browser-`.
- **`-tab` dropped from the seven portal groups** (30 files), matching the prose that now says *views*. The four view overviews are consistent: `portal-{account,api,home,support}-overview`.

Renames need no redirect entries: `scripts/redirects.json` has no asset paths, built asset URLs are content-hashed, and nothing outside `cloud/**/*.mdx` references these files.

## Duplicate and orphan assets

Six basenames existed in both `cloud/assets` and `cloud/portal/assets`, three already drifted apart. Consolidated onto the portal copy and repointed the `cloud/*.mdx` references — which also retires a stale-chrome copy of `portal-product-list-manage-button`. One unreferenced asset deleted.

## Screenshots

- The cluster-health figure had the **Incidents History panel captured twice**; now one panel, with 2026 dates.
- The backup-task figure now shows **Ongoing Tasks**, matching the prose, and evidences the **14-day retention policy** the page claims.
- The tier and instance figures are off the pre-redesign theme, so the product-create Plan set is uniform. **Production tier limits rose to 64 cores / 256 GB RAM / 4000 GB disk**, and the instance cards now carry IOPS and throughput.
- The pricing figures show list rather than discounted prices, and surface **Public IP, backup and per-feature costs** that *Additional Expenses* does not yet cover — see Follow-ups.

## Figure/prose corrections found by re-reading the figures

- **`cloud-scaling`** — the IO-limits admonition said the *Disk* Throughput is lower than the *Instance* Throughput. It is the other way round in the figure, and was in the previous figure too, which inverted the point of the advice that follows it.
- **`cloud-overview`** — the Open Studio walkthrough quoted a certificate name (`"free.serez"`) that no figure has shown for some time; the instruction now points at the URL without quoting an example.
- **`cloud-pricing-payment-billing`** — the prose compared PB10 and **PB20** while the figures are PB10 and **PB30**.
- **Alt text** — the email-notification figure in `cloud-scaling` was labelled *"Scaling storage"*, and two billing figures were labelled `PayNow1` / `PayNow2`.

## Follow-ups, not in this PR

Screenshot values in these figures come from the test environment and are not a concern.

**Stale figure content**

- `portal-product-details-serverless.png` is the last figure on the pre-redesign product view — Manage / Metrics / Cluster Health as buttons, no Alerts tab, and the dead "Instance access" card. It contradicts the refreshed `portal-product-details.png` on the same page.
- `portal-product-create-customize-section.png` shows release channels LTS 5.4 / Stable 6.2 / **Current (7.1)**; `CURRENT_VERSION` is 7.2.
- `studio-cluster-view-with-additional-nodes.png` shows RavenDB 6.2.0.
- `portal-billing-costs-current-billing-period.png` carries 2025 dates beside a September-2026 sibling; `portal-billing-costs-billing-history.png` has no 2026 in its year picker.
- Both `studio-notification-*` figures are dated 14 July 2019; `portal-product-details-audit-logs.png` is December 2024; `portal-api-create-api-key-modal.png` shows an expiration already in the past.

**Figures that show a control but not a result**

- `product-features-powerbi-configuration.png` — IP table is empty, headers only.
- `portal-product-details-cluster-health.png` — *Instance changes in time* is a blank plot area taking roughly 30% of the image height.
- `registration-page-thank-you.png` — roughly 70% empty canvas.

**Prose to check against the figures**

- `cloud-scaling.mdx:133` states Azure **Standard** disks are *500 IOPS and 60 MB/s*, while the adjacent figure lists `500 IOPS / 100 MBps (default)` in the Premium Performance tier dropdown. Same IOPS, different throughput — these may describe different things, but a reader will take them as one claim. The AWS equivalent (`3000 IOPS / 125 MB/s`) does match its figure.
- Three login and registration figures still show the *"Captcha was disabled by configuration option"* banner, a state production never renders.
- The registration walkthrough disagrees with itself across steps: the billing step gives the company as *YasuIT*, the summary says *RavenDB*.

**Content the figures now show and the prose does not**

The refreshed pricing figures surface **Public IP $0.005/hour**, **backup at $1.00/GB/month after the free 1 GB**, and **PowerBi / OLAP ETL / Queue Sink at 10% of hosting cost** — none covered by *Additional Expenses*. The first is what **RDoc-2661** and **RDoc-2662** ask for.

Separately, **verify the hourly prices** in `portal-product-create-plan-pricing-pb10/pb30.png` against current pricing. It is the only remaining item that could be actively wrong for a customer rather than merely dated.

**Near-duplicate pairs the renaming exposed**

Two Studio views each exist as two crops used on two different pages: `studio-server-dashboard-manage-certificates{,-link}.png` and `studio-certificates-{client-certificate-menu,generate-client-certificate}.png`. Each pair is a candidate to collapse, but that is a decision about the pages rather than a rename.

**Other**

- The autoscaling *IO wait min value* field renders empty while the summary beside it shows `0%` — no RDBCL ticket yet, and `product-features-autoscaling-configuration-edit.png` ships the behaviour as if correct.
- 66 images in the marketplace, tiers, backup, migration and studio groups have not been re-reviewed. Every one sampled showed pre-redesign chrome.
- Repo-wide, roughly 898 alt-text tags containing `&` render double-escaped. Pre-existing pipeline behaviour, deliberately out of scope here.


cloud/portal/cloud-portal-billing-tab.mdx
```diff
@@ -43,26 +45,128 @@ The Billing & Costs tab will show your existing resources and their costs as wel
...
+chart.
+
+The available categories are:
+
+| Group | Categories |
+|---|---|
+| Instances and storage | Machine Cost, Storage, Additional Storage, Azure Performance Tier |
+| Traffic and disk | Data Transfer, Disk Operations |
+| Backups | Backup |
+| [Product features](../cloud-features.mdx) | Monitoring, Power BI, OLAP ETL, Queue ETL, Queue Sink, Snowflake ETL, Data Archival, Autoscaling, Serverless Autoscaling, Automatic Storage Extension, AI Embeddings |
```
matz985:
```suggestion
| [Product features](../cloud-features.mdx) | Monitoring, Power BI, OLAP ETL, Queue ETL, Data Archival, AI Embeddings etc. |
```

cloud/portal/cloud-portal-support-tab.mdx
```diff
@@ -35,40 +37,83 @@ A user with the account-level [**Support**](../../cloud/cloud-account.mdx#accoun
...
+Click **Choose plan** on the one you want. An *Upgrade to premium support* dialog opens with a summary of the plan you
+picked, which you confirm with **Send activation request**.
 
-!["Figure 5 - Support Activation Request Form"](./assets/portal-support-activation-request-form.png)
+![Upgrade to premium support: 12 months plan](./assets/portal-support-activation-request-form.png)
+
+![Upgrade to premium support: 6 months plan](./assets/portal-support-activation-request-form-6-months.png)
+
+<Admonition type="note" title="Requirement">
+The paid support option requires at least one **production tier** instance, with a minimum of **$79/month**.
```
matz985: I undestand this sentence as - "at least one production product with minimum cost of  $79/month is required" 
Which is not correct I believe, $79/month is the minimum cost of support, not a threshold for eligibilty.

cloud/portal/cloud-portal.mdx
```diff
@@ -21,37 +21,79 @@ takes you to your account's Portal, which gives you access to all the different
...
+* **[Account](../../cloud/portal/cloud-portal-account-tab.mdx)**  
+  View and edit your invoices and credit card information.
+
+<Admonition type="note" title="">
+Which views you see depends on your access — for example, *Billing & costs* requires the **Billing**
+[account role](../../cloud/cloud-account.mdx#account-roles), and *Users & Access* requires
+**Account with User Management**. Views you have no role for are not shown.
+
+**Products** is the exception: it is not tied to an account role but to your
+[per-product permissions](../../cloud/cloud-account.mdx#product-permissions) — the view appears only if you have access
```
matz985:
```suggestion
Products is gated differently from the other views: it appears if you hold the account-level Products_Create role, or if you have per-product permissions on at least one product. Account owners always see it.
```
more like this I think

matz985 (approved):

PR #2655, opened 2026-09-07
RDoc-4096 - redesign Quill's start page

mateuszbartosik:
### Issue link
https://issues.hibernatingrhinos.com/issue/RDoc-4096

### Additional description

Redesign of the Quill start page (`/quill`), after the feedback that the previous version looked weak.

**Hero.** New `QuillHero` built around `QuillWidgetDemo` — a non-interactive re-creation of the real chat widget that types a question, shows the thinking dots, streams the answer, then cycles through the scripted conversations in `quillDemoConversations.ts`. It replaces the static screenshot. Its palette is the widget's own `--rq-*` tokens for the default theme, so it stays faithful to `packages/widget` rather than inventing a look.

**Channel blob field.** `IntegrationBlobField` floats the channel marks (Telegram, Slack, Discord, chat widget) behind the hero across three depth layers — two crisp in front, the rest progressively blurred and faded — so the hero reads as "many channels" without asking the reader to identify each one. It is decorative only: `aria-hidden`, nothing interactive.

The positioning is worth a note for reviewers. Slot coordinates in `blobPositions.ts` are offsets from the *widget card*, expressed in multiples of its half-width and half-height and measured from live geometry at runtime, not percentages of the hero. The hero is fluid, so percentages could not hold one composition across widths — the blobs collided with the card when narrow and flew to the corners when wide. Anchored to the card, the field hugs the panel at any width. Verified at 800 / 900 / 1100 / 1280 / 1440 / 1920px: every blob lands where the slot table specifies, stays inside the hero, and keeps a visible sliver clear of the card. Below `md` the field is replaced by `ChannelRow`, a plain row of chips.

A development-only assertion warns in the console if a blob could drift so far behind the card that it disappears.

**Channel marks.** The four vendor marks are inline SVG. The Telegram and Discord `viewBox`es are deliberately padded past their artwork: Simple Icons draws both tangent to the box edges (Telegram is a full-bleed disc, `r=12` at `12,12`) and a fraction of a unit beyond, so an exact `0 0 24 24` box clips the rim and they render with flat sides. Both files carry a comment so the padding is not "tidied" away.

**Rest of the page.** `QuillHowItWorks` adds the 4-node data path and the three getting-started cards, followed by the existing Learn / Operate / Build card grids.

**Housekeeping.** The one-item "Quill" breadcrumb is suppressed on `/quill` — it only repeated the sidebar's "Home" entry — reusing the mechanism already in place for `/samples`. The unused gradient banner SVGs and the old hero screenshot are deleted.

### Type of change

- [ ] Content - docs
- [ ] Content - cloud
- [x] Content - Quill
- [ ] Content - guides
- [ ] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs

### Changes in UX/UI
- [ ] No changes in UX/UI
- [x] Changes in UX/UI (include screenshots and description)

The whole start page above the fold is new: animated widget mockup in place of the screenshot, floating channel blobs behind the hero, and the flow strip. Described in full above. Screenshots to follow.
quill


quill/home.mdx
```diff
@@ -2,46 +2,159 @@
 slug: /
 title: Quill Documentation
 sidebar_label: Home
+# Hides the duplicate "Home" entry from the sidebar — the "Quill" category header already links here.
```
Scooletz, line 5: If this means the `Start` at the top we can go with it.

src/components/Quill/QuillWidgetDemo.tsx
```diff
@@ -0,0 +1,451 @@
...
+        (total, message) => total + message.content.length,
+        state.messages.length
+    );
+
+    useLayoutEffect(() => {
+        const feed = feedRef.current;
+        if (feed) {
+            feed.scrollTop = feed.scrollHeight;
+        }
+    }, [scrollSignal]);
```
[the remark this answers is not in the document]
mateuszbartosik: Doesn't apply on React 19 — this repo is on react/react-dom 19.1.0, and the SSR `useLayoutEffect` warning was removed in v19. The string `"useLayoutEffect does nothing on the server"` is absent from `react-dom` entirely, and the server bundles map the hook to a silent no-op (`useLayoutEffect: noop$1`). Rendering this component through `renderToString` in a development build emits no warnings.

Keeping it as a layout effect on purpose: it sets `scrollTop` before paint, so the streaming feed never shows a frame of un-scrolled content. Switching to `useEffect` would reintroduce that flicker.

src/css/custom.css
```diff
@@ -823,66 +871,45 @@ a {
     display: none;
 }
 
-/* Quill start page background. The blueprint grid and the breathing coral glow are lifted from
-   QuillComingSoon.module.css and painted behind the whole article of quill/home.mdx. The page is
+/* Quill start page background. The blueprint grid and the coral glow are lifted from
+   QuillComingSoon.module.css and painted behind the whole article of quill/home.mdx. The glow is
+   static here: an always-on blurred animation repaints the whole page on every frame. The page is
    pinned by the two classes the theme puts on <html>: the plugin's and the doc id's (a docs page's
    `wrapperClassName` frontmatter is not applied anywhere in this theme). Two pseudo-elements under
```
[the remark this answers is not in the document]
mateuszbartosik: Fixed. The block now describes a single `::after` glow and drops the blueprint-grid reference — the grid stayed in `QuillComingSoon.module.css` and isn't reused here.

One correction: the animation sentence wasn't describing a removed effect. "The glow is static here: an always-on blurred animation repaints the whole page on every frame" was the *reason* for keeping it static. It was ambiguous enough to be worth rewording, so it now reads unmistakably as rationale rather than description.

quill/home.mdx
```diff
@@ -2,46 +2,159 @@
...
+        {
+            icon: "notifications",
+            title: "Channels",
+            description: "A chat widget on your site, or a Telegram, Slack, or Discord bot, carries the conversations between your users and your agents.",
+            url: "/quill/overview#channels",
+        },
+    ]}
+/>
+
+## Three steps to a working chat
```
Scooletz: We ditched counting steps in other PR. Also `/quill/getting-started/overview` mentions the other number of steps.

Also, the previous list of signup, getting started, overview, was simpler and not showing the actual steps.

This list is not inclusive, it omits some steps.

I have no actionable feedback here what to amend. I think it's somewhat incorrect.

Scooletz (changes requested):

quill/home.mdx
mateuszbartosik: Yes, the `Start` menu item is our standard for the default page. This ☝️ basically hides the duplicated menu entry.

quill/home.mdx
mateuszbartosik: Let's just remove this section for now.

Scooletz (approved):

PR #2671, opened 2026-09-15
RDoc-4106 Lack of active state in docs subpage select is misleading

poissoncorp:
### Issue link
[RDoc-4106](https://issues.hibernatingrhinos.com/issue/RDoc-4106) Lack of active state in docs subpage select is misleading

### Additional description
The section you are in was removed from the switcher and replaced by a generic "Start" entry, so the list showed every destination except the current one. In Quill that left "Guides" and "Samples" reading as Quill's own.

Keep every section on the list and mark the current one active instead. The entry links to the section landing page, so it also covers what "Start" did.

Both sidebars now render one shared list from getSectionNavItems(), which drops the per-section conditionals and the hover-only "Switch" labels. Desktop and mobile had drifted apart in the meantime: different order, and mobile still used the sparkles icon for Quill.

Was:
image
image

Now:
image
image

image

### Type of change

- [ ] Content - docs
- [ ] Content - cloud
- [ ] Content - Quill
- [ ] Content - guides
- [x] Bug fix
- [ ] Optimization
- [ ] Other

### Changes in docs URLs

- [x] No changes in docs URLs

### Changes in UX/UI
- [ ] No changes in UX/UI
- [x] Changes in UX/UI (include screenshots and description)


src/typescript/pathUtils.ts
```diff
@@ -28,3 +30,44 @@ export function getLandingPagePath(pathType: PathTypeValue, versionLabel: string
...
+    label: string;
+    icon: IconName;
+    to: string;
+    pathType: PathTypeValue | null;
+    // Shown only while you are already inside it, for areas that are not public destinations.
+    currentOnly?: boolean;
+    external?: boolean;
+}
+
+const NAV_SECTIONS: readonly { type: PathTypeValue; label: string; icon: IconName; currentOnly?: boolean }[] = [
```
Lwiel, line 44: Can we use a concrete type here?

src/typescript/pathUtils.ts
Lwiel, line 16: Here we can also use a concrete type to be consistent

mateuszbartosik (changes requested):
Please revert the Start menu item, and treat it as home for the product documentation. Adding an active state for menu is fine.

